PR #1, opened 2021-11-08
refactor terraform and makefiles

samwachspress:


tf/admin/cloud-storage.tf
```diff
@@ -0,0 +1,9 @@
+resource "google_storage_bucket" "terraform_state" {
+  name          = "${var.GOOGLE_CLOUD_TF_STATE_BUCKET_NAME}"
+  location      = var.GOOGLE_CLOUD_REGION
+  force_destroy = true
```
r-vieira, line 4: During development this works, but consider setting `force_destroy` to false once this is ready for production.

samwachspress: good catch - i copied this from the src-archive bucket where cloud function code is stored, which i do actually want to force_destroy

tf/caendr/modules/site/cloud-sql.tf
```diff
@@ -0,0 +1,72 @@
...
+  instance  = google_sql_database_instance.postgres_instance.name
+  name      = var.module_site_vars.postgres_db_name
+  project   = var.google_cloud_vars.project_id
+}
+
+resource "google_sql_database" "postgres_db_stage" {
+  charset   = "UTF8"
+  collation = "en_US.UTF8"
+  instance  = google_sql_database_instance.postgres_instance.name
+  name      = var.module_site_vars.postgres_db_stage_name
```
r-vieira, line 13: help me understand the naming here -- assuming the `cloud-sql` module is environment agnostic, are we defining two instances (postgres_dn_name and postgres_db_stage_name). Does `_stage` have a special meaning within a given environment?

tf/caendr/provider.tf
```diff
@@ -0,0 +1,26 @@
+terraform {
+  backend "gcs" { }
+  required_version = ">= 1.0.0"
```
r-vieira, line 3: consider limiting on the upper-bound, eg: version = ">= 1.0.0, < 2.0.0"

r-vieira (approved):
Looking good. Minor comments inline.

tf/caendr/modules/site/cloud-sql.tf
samwachspress: Updating the DB with update-or-replace queries can have issues when the external data sources being pulled have changes that break relationships in the data. Dan was just dropping the tables and then rebuilding them from scratch, but that has significant downtime, temporarily incomplete data, and if the task fails it needs to be rolled back. The stage DB is used for all the ETL operations, then dumped as a .sql backup in cloud storage and dropped. That .sql can be loaded much faster to the active db or the stage db can be used while the master db is being rebuilt. Since the only billing is for the instance and not the db, I figured this approach would reduce downtime without adding cost.

PR #13, opened 2022-02-11
Added new styles for homepage.

natalieroman:
Modified container-fluid width globally.
Added service card images.
Created new default template for homepage.
Fixed html layout issues on homepage, genome browser, variant annotation pages.
Updated footer and navbar includes to support new homepage design.


natalieroman:
There may be issues with image paths for the service cards and the CeNDR logo in the navbar.

PR #55, opened 2022-03-16
Feature/password reset

gmg300:


src/modules/site/base/forms/forms.py
```diff
@@ -76,6 +76,16 @@ class BasicLoginForm(FlaskForm):
   password = PasswordField('Password', [Required(), Length(min=5, max=30)])
   recaptcha = RecaptchaField()
 
+class PasswordResetForm(FlaskForm):
+  """ The password reset form """
+  password = PasswordField('New Password', [Required(), EqualTo('confirm_password', message='Passwords must match'), Length(min=5, max=30)])
```
r-vieira, line 81: bump min to at least 12

src/modules/site/base/forms/forms.py
```diff
@@ -76,6 +76,16 @@ class BasicLoginForm(FlaskForm):
...
 
+class PasswordResetForm(FlaskForm):
+  """ The password reset form """
+  password = PasswordField('New Password', [Required(), EqualTo('confirm_password', message='Passwords must match'), Length(min=5, max=30)])
+  confirm_password = PasswordField('Confirm New Password', [Required(), EqualTo('password', message='Passwords must match'), Length(min=5, max=30)])
+  recaptcha = RecaptchaField()
+
+class RecoverUserForm(FlaskForm):
+  """ The account recovery email form """
+  email = EmailField('Email Address', [Required(), Email(), Length(min=6, max=50)])
```
r-vieira, line 87: bump max to 320 (confirm in the spec for accuracy)

src/modules/site/base/utils/auth.py
```diff
@@ -15,11 +20,22 @@
...
-                                jwt_required)
+                                jwt_required,
+                                decode_token)
 
 from caendr.models.datastore import User
+from caendr.models.datastore.user_token import UserToken
 from extensions import jwt
 
+def create_one_time_token(id, destination='/'):
+  expires_delta = timedelta(seconds=JWT_LIMITED_TOKEN_EXPIRES)
```
r-vieira, line 31: consider prefixing the ENV name with 'PASSWORD_RESET_' to avoid confusing it with the regular JWT session token

src/modules/site/base/utils/auth.py
```diff
@@ -15,11 +20,22 @@
...
+  token = create_access_token(identity=str(id), additional_claims={'destination': destination}, expires_delta=expires_delta)
+  decoded_token = decode_token(token)
+  jti = decoded_token['jti']
+  user_token = UserToken(jti)
+  user_token.set_properties(username=id, revoked=False)
```
r-vieira, line 36: hum... do we need the field `revoked`? I was wondering if the mere existence of an entity in the datastore with kind=`user_expired_token` would suffice for the check. Or reframing the question, is the `revoked` field ever not `False`?

src/modules/site/base/utils/auth.py
```diff
@@ -81,9 +97,22 @@ def unauthorized_callback(reason):
...
+    logger.info(f"Revoking user token - {jti} for user - {user_token.username}")
+    user_token.revoke()
+    return token_revoked
+    
+
 @jwt.invalid_token_loader
 def invalid_token_callback(callback):
   ''' Invalid Fresh/Non-Fresh Access token in auth header, redirect to login '''
+  logger.info("Invalid Token.")
   resp = make_response(redirect(url_for('auth.choose_login')))
```
r-vieira, line 116: consider adding a 'flash'/banner explaining what happened to the token to the user.

src/modules/site/base/views/user.py
```diff
@@ -41,6 +45,39 @@ def user_register():
...
+    if users and users[0]:
+      user = users[0]
+      token = create_one_time_token(id=user.get('username'), destination=url_for('user.user_reset_password'))
+      password_reset_link = url_for('auth.token', token=token, _external=True)
+      try:
+        send_email({
+          "from": "no-reply@elegansvariation.org",
+          "to": [ email ],
+          "subject": "CeNDR Password Reset",
+          "text": PASSWORD_RESET_EMAIL.format(email=email, password_reset_link=password_reset_link)
```
r-vieira, line 68: consider adding suffix `_TEMPLATE` to the `PASSWORD_RESET_EMAIL`

src/modules/site/base/views/user.py
```diff
@@ -68,3 +105,22 @@ def user_update():
...
+@user_bp.route("/password/reset", methods=["GET", "POST"])
+@jwt_required()
+def user_reset_password():
+  title = "Reset Password"
+  disable_parent_breadcrumb = True
+  user = get_current_user()
+  form = PasswordResetForm(request.form)
+
+  if request.method == 'POST' and form.validate():
+    password = slugify(request.form.get("password"))
```
r-vieira, line 119: does the form.validation assert that password and password-verify match on the server-side?

src/modules/site/constants.py
```diff
@@ -7,6 +7,7 @@ class PRICES:
   STRAIN = 15
   SHIPPING = 65
 
+JWT_LIMITED_TOKEN_EXPIRES = 15 * 60
```
r-vieira, line 10: Consider prefixing with `PASSWORD_RESET_` and moving this to the `global.env` under `env/main/global.env`. Add it to the `site` section of `global.env`.

src/pkg/caendr/caendr/models/datastore/user_token.py
```diff
@@ -0,0 +1,29 @@
+from caendr.models.datastore import Entity
+
+
+class UserToken(Entity):
```
r-vieira, line 4: consider renaming to `UserRevokedToken`

src/pkg/caendr/caendr/services/cloud/datastore.py
```diff
@@ -20,7 +20,7 @@ def get_ds_entity(kind, name):
     for k, v in result.items():
       if isinstance(v, str) and v.startswith("JSON:"):
         result_out[k] = json.loads(v[5:])
-      elif v:
+      elif v or isinstance(v, bool):
```
r-vieira, line 23: nice catch!

r-vieira (changes requested):

src/modules/site/base/views/user.py
gmg300: Yes, it applies the any validators set when creating the FlaskForm on the form data that was posted.

src/pkg/caendr/caendr/models/datastore/user_token.py
gmg300: I'm going to leave the current name for now since not all tokens of that DS kind are revoked.

src/modules/site/base/views/user.py
```diff
@@ -41,6 +45,39 @@ def user_register():
...
+      try:
+        send_email({
+          "from": "no-reply@elegansvariation.org",
+          "to": [ email ],
+          "subject": "CeNDR Password Reset",
+          "text": PASSWORD_RESET_EMAIL_TEMPLATE.format(email=email, password_reset_link=password_reset_link)
+        })
+        logger.info(f"Sent password reset email: {email}, link: {password_reset_link}")
+      except Exception as err:
+        logger.error(f"Failed to send email: {err}")
```
r-vieira, line 72: let's discuss this -- it appears that you are trying to avoid leaking any details if an email exists in the system, which is an excellent best practice. 
In this particular case, the email subsystem may actually throw an error (e.g. invalid API, etc), and that should be communicated to the user with something along the lines of "Unable to reset your credentials at this time. Please try again later." as a flash message.

src/modules/site/base/views/user.py
```diff
@@ -41,6 +45,39 @@ def user_register():
...
+    
+    # Flash info message even if email fails in order to protect from attempts at finding valid emails
+    flash(f"Please check your email ({email}) for a password reset link.", 'info')
```
r-vieira, line 75: If an exception was caught, on the email sending, we should not notify the users to expect an email. Consider a slightly different workflow that allows for all the code paths to yield the best 'hints' for the end-user.

src/modules/site/base/views/user.py
```diff
@@ -68,3 +105,22 @@ def user_update():
...
+
+@user_bp.route("/password/reset", methods=["GET", "POST"])
+@jwt_required()
+def user_reset_password():
+  title = "Reset Password"
+  disable_parent_breadcrumb = True
+  user = get_current_user()
+  form = PasswordResetForm(request.form)
+
+  if request.method == 'POST' and form.validate():
```
r-vieira, line 118: The more I see this pattern, the more I want it to read.

```
if not (request.method == "POST" and form.validate()):
    return # render template/etc.
# (...)
return redirect(...)
```
Thinking of ways to codify this syntax preference for if statements -- taking a stab at it: _always prefer the version of the code with the least amount of indentation levels (within reason) that achieve the same behavior._

r-vieira (changes requested):
Great work!! See inline comments

src/modules/site/base/views/auth/auth.py
```diff
@@ -20,21 +20,42 @@
...
-                    __name__,
-                    template_folder='templates')
+auth_bp = Blueprint('auth', __name__, template_folder='templates')
 
 @auth_bp.route('/')
 def auth():
   return redirect(url_for('auth.choose_login'))
 
+@auth_bp.route('/token')
+def token():
```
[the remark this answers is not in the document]
gmg300: My thinking for this route was 3 checks that short circuit and re-route to basic login:

1. Check for token existence, validity and expiration
2. Check if token was revoked
3. Check that user associated with token exists

Do you think we can/should combine two of these checks?

When running the 'decode_token()' a couple callbacks can be triggered if the token is expired or invalid which then redirects to login. Those callbacks don't include a check if the token is revoked, that has to be called manually since this route doesn't include @jwt_required.

src/modules/site/base/views/user.py
gmg300: In other words, always prefer to have less code indented if possible?

The problem I see though is that we don't have a fall through if we short circuit if some cases like this one:
```
@user_bp.route("/recover", methods=["GET", "POST"])
def user_recover():
  """ Enter an email address for a user to send a password reset link """
  title = "Recover User"
  disable_parent_breadcrumb = True
  form = RecoverUserForm(request.form)

  if request.method == 'POST' and form.validate():
    email = request.form.get("email")
    users = get_local_user_by_email(email)

    if not users or not users[0]:
      # Flash info message even if email fails in order to protect from attempts at finding valid emails
      flash(f"Please check your email ({email}) for a password reset link.", 'info')
      logger.error(f"Failed to retrieve LOCAL user for email: {email}")
      return redirect('/')
    
    user = users[0]
    token = create_one_time_token(id=user.get('username'), destination=url_for('user.user_reset_password'))
    password_reset_link = url_for('auth.token', token=token, _external=True)
    try:
      send_email({
        "from": "no-reply@elegansvariation.org",
        "to": [ email ],
        "subject": "CeNDR Password Reset",
        "text": PASSWORD_RESET_EMAIL_TEMPLATE.format(email=email, password_reset_link=password_reset_link)
      })
      logger.info(f"Sent password reset email: {email}, link: {password_reset_link}")
    except Exception as err:
      flash('Unable to reset your credentials at this time. Please try again later.', 'error')
      logger.error(f"Failed to send email: {err}")
    
  return render_template('user/recover_user.html', **locals())
```

src/modules/site/base/forms/forms.py
```diff
@@ -76,6 +76,16 @@ class BasicLoginForm(FlaskForm):
   password = PasswordField('Password', [Required(), Length(min=5, max=30)])
   recaptcha = RecaptchaField()
 
+class PasswordResetForm(FlaskForm):
+  """ The password reset form """
+  password = PasswordField('New Password', [Required(), EqualTo('confirm_password', message='Passwords must match'), Length(min=12, max=30)])
+  confirm_password = PasswordField('Confirm New Password', [Required(), EqualTo('password', message='Passwords must match'), Length(min=12, max=30)])
```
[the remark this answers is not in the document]
gmg300: That Length validator comes from wtforms.validators. Does it make sense to move that into the config file?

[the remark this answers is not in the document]
r-vieira: to keep the implementation decoupled from settings, we may want to keep a set of those rules in the config, eg:
```
PASSWORD_MIN_LENGTH=N
PASSWORD_MAX_LENGTH=N
```
and then use with `Length(min=PASSWORD_MIN_LENGTH,...)`

src/modules/site/base/views/auth/auth.py
```diff
@@ -20,21 +20,42 @@
...
+    decoded_token = decode_token(token)
+    token_revoked = check_if_token_revoked(decoded_token)
+
+    if not token_revoked:
+      username = decoded_token.get('sub')
+      user = User(username)
+      destination = decoded_token.get('destination')
+        
+      if user._exists:
+        user.set_properties(last_login=datetime.now(timezone.utc))
```
r-vieira, line 51: this is a good candidate to be a method on the `User` class. eg: `User.updateLastLoginTime()`. This makes the `auth.py` not have to know anything about internals for the `User` class.

r-vieira (approved):

PR #56, opened 2022-03-17
 bugfix/vbrowser gene search

gmg300:
Adding filter to process_gene_search results to show only items that include all values needed to calculate position


src/modules/site/templates/data/vbrowser.html
```diff
@@ -294,9 +294,10 @@ <h3 class="text-center">Variant Impact</h3>
          url: "{{ url_for('api_gene.api_search_combined', query='') }}" + gene,
          method: "GET",
          contentType: 'application/xml',
-         }).done(function(msg) {
+         }).done(function(results) {
           row = Array();
```
r-vieira, line 298: is `row` being used globally elsewhere? If so, we should refactor, otherwise, consider removing it here.

r-vieira (changes requested):

src/modules/site/templates/data/vbrowser.html
```diff
@@ -393,16 +392,16 @@ <h3 class="text-center">Variant Impact</h3>
 function renderTargetConsequenceTable(data) {
   let result = "<div class='instruction-well'><table class='table table-striped table-condensed table-hover' style='width:80%;margin-left:10%;font-size:0.8rem;'><thead>";
   const numRows = Object.keys(data.id).length;
-  for (i of columnList) {
-    result += `<th>${i.name}</th>`;
+  for (col1 of columnList) {
```
r-vieira, line 395: narrow the scope of `col1` -- add `let col1....`

src/modules/site/templates/data/vbrowser.html
```diff
@@ -393,16 +392,16 @@ <h3 class="text-center">Variant Impact</h3>
...
-    result += `<th>${i.name}</th>`;
+  for (col1 of columnList) {
+    result += `<th>${col1.name}</th>`;
   }
   result += "</thead><tbody>";
   for (let j = 0; j < numRows; j++) { 
     result += "<tr>";
-    for (k of columnList) {
-      let val = data[k.id][j] || "";
+    for (col2 of columnList) {
```
r-vieira, line 401: narrow scope of `col2`

src/modules/site/templates/data/vbrowser.html
```diff
@@ -491,45 +479,46 @@ <h3 class="text-center">Variant Impact</h3>
   let result = '<mark></mark>';
   let strains = data.split(',');
   let i = 0;
-  for(s of strains) {
+  for(strain of strains) {
```
r-vieira, line 482: narrow scope for `strain`

src/modules/site/templates/data/vbrowser.html
```diff
@@ -491,45 +479,46 @@ <h3 class="text-center">Variant Impact</h3>
...
+      case 'target_consequence':
+        return { 
+          "name": col.id, 
+          "mRender": renderTargetConsequenceCol,
+          "targets": idx
+        } 
+      default:
+        return { 
+          "name": col.id, 
+          "targets": idx
```
r-vieira, line 510: can you keep the shape of the returned value? ie. assign the `mRender` to a no-op-function like `nop = () => ''`.

src/modules/site/templates/data/vbrowser.html
```diff
@@ -625,7 +614,7 @@ <h3 class="text-center">Variant Impact</h3>
       $(".toggle-col-chk").each(function(i, e) {
         e.checked = state;
       });
-      for (let i = 0; i < {{ columns.__len__() }}; i++) {
+      for (let i = 0; i < columnList.length; i++) {
```
r-vieira, line 617: narrow scope of `i`

r-vieira (changes requested):

src/modules/site/templates/data/vbrowser.html
gmg300: It looks like the mRender value (now called render in datatables 1.10) has to be null or undefined. It'll consider it valid if it's type function and will try to execute.

I refactored to retain the shape of the return value and just get the render value based on the col.id

src/modules/site/templates/data/vbrowser.html
```diff
@@ -276,11 +276,10 @@ <h3 class="text-center">Variant Impact</h3>
   downloadCSV(jsonData);
 }
 
+const strainListing = JSON.parse('{{ strain_listing | tojson }}');
 
 function init_selected_strain_set() {
-{% for strain in strain_listing %}
-  selected_strains.add('{{strain}}');
-{% endfor %}
+  strainListing.forEach(strain => selected_strains.add(strain));
```
r-vieira, line 282: unless there are filters/side-effects, consider `selected_strains = new Set(strainListing);`

r-vieira (approved):

PR #83, opened 2022-04-06
feature/view all results

gmg300:
Creating #get_all... methods for mapping, heritability and indel_primer, Refactoring results/reports into 'all' and 'user' routes for mapping, heritability and indel_primer, Refactoring list template for mapping, heritability and indel_primer into list-all and list-user templates, Refactoring dataTable_top_menu to remove buttons that are set as 'admin_only' when user is not admin, Adding buttons to mapping, heritability and indel_primer data tables to redirect to all or user results or #create route

### DIFF
--- a/src/modules/site/templates/_includes/macros.html
+++ b/src/modules/site/templates/_includes/macros.html
@@ -77,12 +77,16 @@
   </div> {# /col-md-2 #}
   <div class="col-md-6">
     {% for action in actions %}
-      <a 
-        class="{{ action.get('classes', '') }}"
-        style="{{ action.get('style','') }}"
-        href="{{ url_for(action.get('url')) }}"
-      > {{ action.get('label') }}
-      </a>
+      {% if not action.get('admin_only', False) or (action.get('admin_only', False) and session['is_admin']) %}
+        <a 
+          class="{{ action.get('classes', '') }}"
+          style="{{ action.get('style','') }}"
+          href="{{ action.get('url') }}"
+        >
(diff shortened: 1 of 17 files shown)


src/modules/site/templates/_includes/macros.html
```diff
@@ -77,12 +77,16 @@
   </div> {# /col-md-2 #}
   <div class="col-md-6">
     {% for action in actions %}
-      <a 
-        class="{{ action.get('classes', '') }}"
-        style="{{ action.get('style','') }}"
-        href="{{ url_for(action.get('url')) }}"
-      > {{ action.get('label') }}
-      </a>
+      {% if not action.get('admin_only', False) or (action.get('admin_only', False) and session['is_admin']) %}
```
r-vieira, line 80: you can use session.get('is_admin') which won't throw a KeyError, and return None instead if the key does not exist.

src/modules/site/templates/tools/heritability/list-all.html
```diff
@@ -0,0 +1,117 @@
+{% extends "_layouts/default.html" %}
+
+{% block custom_head %}
+    <link rel="stylesheet" href="//cdn.datatables.net/1.10.21/css/jquery.dataTables.min.css" />
```
r-vieira, line 4: we are loading a different version of databables JS + CSS in `head.html`. Can you remove this dependency here and use the global? Otherwise, if you need this version specifically we may need to refactor the global one into the templates where it is actually used.

r-vieira (approved):

PR #92, opened 2022-04-11
feature/sentry init frontend

gmg300:
Adding and initializing Sentry in template head for frontend


gmg300:
@r-vieira Should I maybe expose the environment in src/modules/site/application.py with a function like #get_env() and then pass that into the Sentry.init() in the template head.html?

PR #127, opened 2022-11-07
Fix variant impact filtering

vlagrassa:
Fix to Variant Annotation tool.

Set table filtering to match check boxes from the start:
- Draw initial values from the checkboxes in HTML
- Filter by impact at the beginning, to initialize properly


r-vieira (approved):
lgtm

PR #133, opened 2022-11-23
Move protocols.yaml to GCP

vlagrassa:


src/modules/site/base/views/strains.py
```diff
@@ -159,7 +160,17 @@ def strains_submission_page():
 @strains_bp.route("/protocols")
 @cache.memoize(60*60)
 def protocols():
-    title = "Protocols"
-    protocols = yaml.safe_load(
-        open("base/static/yaml/protocols.yaml", 'r'))
-    return render_template('strain/protocols.html', **locals())
+
+    # Get location of YAML file in GCP
+    MODULE_SITE_BUCKET_ASSETS_NAME = os.environ.get('MODULE_SITE_BUCKET_ASSETS_NAME')
```
r-vieira, line 165: hum...i wonder if the `protocols.yml` should live in the same bucket where the data releases are hosted (caendr-site-public-bucket). Any pros/cons?

r-vieira (changes requested):
Looks good -- see note about which bucket should host the `protocols.yml`

src/modules/site/base/views/strains.py
vlagrassa: Pros:
- It's where all the other 'protocols' files are located (as far as I can tell), so it keeps the whole page in one place

Cons:
- I suppose it's technically not 'static', as the name `caendr-site-static-bucket` would imply? If we want to allow admins to edit this file from the webpage, it would make sense to put it elsewhere

r-vieira: Also, it is possible that someone [external] might have linked directly to one of the PDFs under the static bucket, since that is the link shown on the website. Let's keep it under the static.

r-vieira (commented):
approved

PR #142, opened 2023-02-03
New Indel Primer filename format (hardcoded env vars)

vlagrassa:
Swap out the names of the Indel Primer files to the new format, so the most recent data is available on the site.

Filenames still hardcoded as environment variabless - pending update to site for multiple species, update to DB ops to construct tables from correct sources, and changes to Indel Primer tool to construct URLs from species + release.


r-vieira (approved):
lgtm

PR #145, opened 2023-02-16
Bring merged feature branch into new design branch

vlagrassa:
Still TODO:
- Update gbrowser and vbrowser HTML sections with strain lists
- Update all tools with species list
- Hook up API calls more broadly


src/modules/site-v2/base/views/about/about.py
```diff
@@ -31,7 +31,7 @@ def about():
   ''' About us Page - Gives an overview of CaeNDR '''
   title = "About CaeNDR"
   disable_parent_breadcrumb = True
-  isotypes = get_isotypes(known_origin=True)
+  isotypes = get_isotypes(known_origin=True, species='c_elegans')
```
r-vieira, line 34: @vlagrassa -- based on our earlier conversation, let's move towards removing all strings as identifiers from the code base and instead use "enums" or class-enums. See the caendr.model for `TaskStatus` as an example.

src/modules/site-v2/base/views/api/gene.py
```diff
@@ -25,16 +26,18 @@ def api_search_homologs(query=""):
...
-
-
-@api_gene_bp.route('/search/<string:query>')
-@cache.memoize(60*60)
-@jsonify_request
-def api_search_combined(query=""):
-  query = request.args.get('query') or query
-  query = str(query).lower()
-  return (search_genes(query) + search_homologs(query))[0:10]
+  species = request.args.get('species') or None
```
r-vieira, line 29: The `#get() + or` can be combined: the 2nd argument to #get(first, second) is the default value. Eg: 
```
a = {}
b = a.get('foo', 'bar')
b == 'bar' # True
```

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -22,45 +30,79 @@ def genetic_mapping():
...
+    # Page info
+    'title': 'Genetic Mapping',
+    'alt_parent_breadcrumb': {"title": "Tools", "url": url_for('tools.tools')},
+
+    # Form info
+    'jwt_csrf_token': (get_jwt() or {}).get("csrf"),
+    'form': FileUploadForm(),
+
+    # TODO: change
+    'nemascan_container_url': 'https://github.com/AndersenLab/dockerfile/tree/nemarun/nemarun',
```
r-vieira, line 44: @vlagrassa we track the container info in DataStore. It might be helpful to add the 'git repo' as a new property to the container image and reference it here.

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -22,45 +30,79 @@ def genetic_mapping():
...
-  
-  label = bleach.clean(request.form.get('label'))
-  props = {'label': label, 
-          'username': user.name,
-          'email': user.email,
-          'file': request.files['file']}          
+
+  # Read fields from form
+  label   = bleach.clean(request.form.get('label'))
+  species = bleach.clean(request.form.get('species'))
```
r-vieira, line 71: add sanitization & validation to all user input -- handle the cases when `label` and `species` are not defined or truncates to empty string, etc.

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -22,45 +30,79 @@ def genetic_mapping():
...
+
+  # Save uploaded file to server temporarily with unique generated name
+  # TODO: Is there a better way to generate this name? E.g. using a Tempfile?
+  local_path = os.path.join(uploads_dir, secure_filename(f'{ unique_id() }.tsv'))
+  request.files['file'].save(local_path)
```
r-vieira, line 76: handle failed `save` operation and notify the user.

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -94,7 +136,7 @@ def report(id):
   fluid_container = True
   data_url = generate_blob_url(mapping.get_bucket_name(), mapping.get_data_blob_path())
   # if hasattr(mapping, 'mapping_report_url'):
-  if hasattr(mapping, 'report_path'):
+  if mapping.report_path is not None:
```
r-vieira, line 139: Handle mapping not found when user submits an invalid <id>

src/modules/site-v2/base/views/tools/genome_browser.py
```diff
@@ -44,36 +94,70 @@ def get_dataset_release_or_latest(release_version = None):
...
-          },
-          'wormbase_version': wormbase_version,
-          'release_version': int(dataset_release.version),
-          'track_url_prefix': track_url_prefix,
-          'bam_bai_url_prefix': bam_bai_url_prefix,
-          'dataset_release_prefix': dataset_release_prefix,
-          'fluid_container': True}
-  return render_template('tools/genome_browser/gbrowser.html', **VARS)
+    # wormbase_version = wormbase_version_override or dataset_release.wormbase_version
+    wormbase_version = WormbaseVersion('WS276')
```
r-vieira, line 105: move hardcoded value `WS276` to either datastore or external configuration object.

src/modules/site/base/forms/forms.py
```diff
@@ -192,16 +192,26 @@ def process_formdata(self, val):
...
+
+
 class PairwiseIndelForm(Form):
-  STRAIN_CHOICES = get_indel_primer_strain_choices()
   CHROMOSOME_CHOICES = get_indel_primer_chrom_choices()
+  SPECIES_CHOICES = [(name, value.short_name) for name, value in SPECIES_LIST.items()]
   
-  strain_1 = SelectField('Strain 1', choices=STRAIN_CHOICES, default="N2", validators=[Required(), validate_uniq_strains])
-  strain_2 = SelectField('Strain 2', choices=STRAIN_CHOICES, default="CB4856", validators=[Required()])
+  species = SelectField('Species', choices=SPECIES_CHOICES, default="c_elegans", validators=[Required()])
```
r-vieira, line 204: @vlagrassa can we move these default settings under the "Tool Version/Settings" in the admin interface? We can store these under datastore/etc. (goal is to to remove all hard-coded vars from code).

src/modules/site/base/views/about.py
```diff
@@ -40,7 +40,7 @@ def about():
   ''' About us Page - Gives an overview of CeNDR '''
   title = "About CeNDR"
   disable_parent_breadcrumb = True
-  isotypes = get_isotypes(known_origin=True)
+  isotypes = get_isotypes(known_origin=True, species='c_elegans')
```
r-vieira, line 43: Avoid hardcoded strings -- see previous comment on creating a class/enum to represent the species.  (eg: models.TaskStatus).

src/modules/site/base/views/mapping.py
```diff
@@ -21,44 +29,75 @@ def mapping():
...
+    'label':    label,
+    'species':  species,
+    'filepath': local_path,
+  }
+
   try:
-    logger.warn("Check Duplicates is DISABLED !!!")
-    m = create_new_mapping(**props, check_duplicates=True)
+    # logger.warn("Check Duplicates is DISABLED !!!")
+    m = create_new_mapping(user, data, no_cache=False)
```
r-vieira, line 82: careful to cleanup -- we shouldn't return from this function without cleaning up the uploads_dir folder (delete the file at `local_path`) on all the paths for this function on exit.

src/modules/site/base/views/mapping.py
```diff
@@ -21,44 +29,75 @@ def mapping():
...
   except Exception as ex:
-    if str(type(ex).__name__) == 'DuplicateDataError':
-      flash('It looks like you submitted that data already - redirecting to your list of Mapping Reports', 'danger')
-      return redirect(url_for('mapping.mapping_user_reports'))
-    if str(type(ex).__name__) == 'CachedDataError':
-      flash('It looks like that data has already been submitted - redirecting to the saved results', 'danger')
-      return redirect(url_for('mapping.mapping_report', id=ex.description))
-    logger.error(ex.description)
-    flash(f"Unable to submit your request: \"{ex.description}\"", 'danger')
+    logger.error(ex)
```
r-vieira, line 94: can we combine the output of these three logger calls into one?

src/modules/site/base/views/mapping.py
```diff
@@ -89,7 +128,7 @@ def mapping_report(id):
   fluid_container = True
   data_url = generate_blob_url(mapping.get_bucket_name(), mapping.get_data_blob_path())
   # if hasattr(mapping, 'mapping_report_url'):
-  if hasattr(mapping, 'report_path'):
+  if mapping.report_path is not None:
```
r-vieira, line 131: careful -- mapping might be invalid if the <id> was invalid or didn't match a correct mapping.

src/pkg/caendr/caendr/models/datastore/browser_track.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class BrowserTrack(Entity):
+
+
+  ## Initialization ##
+
+  def __new__(cls, *args, **kwargs):
+    if cls is BrowserTrack:
+      raise TypeError(f"Class '{cls.__name__}' should never be instantiated directly -- subclasses should be used instead.")
```
r-vieira, line 19: 👍

src/pkg/caendr/caendr/models/datastore/nemascan_mapping.py
```diff
@@ -1,69 +1,125 @@
...
+    ])
+
+    status = None
+
+    # Loop through all submissions by different users
+    for match in matches:
+      if match.username != self.username:
+
+        # Update to match status, keeping 'COMPLETE' if it's found
+        if status != 'COMPLETE':
```
r-vieira, line 121: replace 'COMPLETE' with TaskStatus.COMPLETE

src/pkg/caendr/caendr/models/datastore/wormbase.py
```diff
@@ -0,0 +1,145 @@
...
+class WormbaseVersion():
+
+    def __init__(self, value):
+
+        if value is None:
+            self._value = None
+
+        # Copy value of another Wormbase Version object
+        elif isinstance(value, WormbaseVersion):
+            self._value = value._value
```
r-vieira, line 18: is value._value guaranteed to be == value.upper()

src/pkg/caendr/caendr/models/datastore/wormbase.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Validate value field of WormbaseProjectNumber class
+        elif isinstance(value, WormbaseProjectNumber):
+            return WormbaseProjectNumber.validate(value._value)
+
+        # Check string value against RegEx
+        return isinstance(value, str) and bool(re.match(r'^PRJNA[0-9]+$', value, flags=re.IGNORECASE))
+
+
+# Test code
```
r-vieira, line 118: @vlagrassa Excellent candidate for a unit test! Take a look at https://docs.python.org/3/library/unittest.html and consider writing a unit test file with these cases. 

Note: look into python unittest discover (it can find your test cases and run them).

src/pkg/caendr/caendr/services/gene_browser_tracks.py
```diff
@@ -4,66 +4,51 @@
...
-  t = GeneBrowserTracks(id)
-  if task:
-    t.set_properties(status='SUBMITTED')
-  else:
-    t.set_properties(status='ERROR')
+  task   = GeneBrowserTracksTask(t)
+  result = task.submit()
+
+  # Update entity status to reflect whether task was submitted successfully
+  t.status = 'SUBMITTED' if result else 'ERROR'
```
r-vieira, line 46: use models.TaskStatus for SUBMITTED and ERROR

src/pkg/caendr/caendr/services/indel_primer.py
```diff
@@ -45,13 +27,6 @@
 MIN_SV_SIZE = 50
```
r-vieira, line 27: can we move these values (MIN_SV_SIZE, MAX_SV_SIZE) under the admin Tool Versions/Settings?

src/pkg/caendr/caendr/services/sql/dataset/_templates.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+
+## Filenames ##
+
+def get_download_path(self, species_name: str = ''):
+  if species_name == '':
+    return self.local_download_path
+  else:
+    return f'{self.local_download_path}/{species_name}'
```
r-vieira, line 119: careful -- you need to validate and *only* accept known species_name. Malicious input might set species_name= '../../../../' to access unauthorized locations.

src/pkg/caendr/caendr/services/sql/dataset/_env.py
```diff
@@ -0,0 +1,39 @@
+import os
+
+from caendr.utils.data import remove_env_escape_chars
+
+
+# Get environment variables
+MODULE_DB_OPERATIONS_BUCKET_NAME = os.environ.get('MODULE_DB_OPERATIONS_BUCKET_NAME')
+STRAIN_VARIANT_ANNOTATION_PATH   = os.environ.get('STRAIN_VARIANT_ANNOTATION_PATH')
+
```
r-vieira, line 9: add guards here -- raise exception if neither of these exists

r-vieira (approved):
See inline comments for the next PR (look for comments with 'careful').

src/modules/site-v2/base/static/js/strains.js
```diff
@@ -0,0 +1,50 @@
...
+
+    // Otherwise, show all strains matching given species
+    else {
+      $(`.strain-entry-container-${ species_id }`).show()
+    }
+  }
+}
+
+
+function update_strain_dropdown(id, strains, species, initial_value = null) {
```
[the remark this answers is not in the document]
vlagrassa: Fixed in fdc46c72f9411734c189cd5ea3f27f3674e806b3

src/modules/site-v2/base/static/js/strains.js
```diff
@@ -0,0 +1,50 @@
...
+
+  // Clear the specified dropdown
+  $(id).empty();
```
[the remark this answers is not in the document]
vlagrassa: Fixed in fdc46c72f9411734c189cd5ea3f27f3674e806b3

src/modules/site-v2/base/views/api/gene.py
vlagrassa: Fixed in 9e2ecf8e2deb2a16724920837796f8f61ccd0b7a

src/modules/site/base/views/mapping.py
vlagrassa: Good catch -- I was performing this inside the `create_new_mapping` function, but A) I think it makes more sense to move up to the view function, and B) I haven't added this for heritability yet.

https://github.com/AndersenLab/CAENDR/blob/feature/multiple-species-etl-ops/src/pkg/caendr/caendr/services/nemascan_mapping.py#L63
```
finally:
    try:
      os.remove(data['filepath'])
    except FileNotFoundError:
      pass
```

vlagrassa: Fixed in 9f68c1cd0eafd697e77e8098cf4ded68852973ab

src/modules/site/base/views/mapping.py
vlagrassa: Fixed in 8ef543d1e139ac211e468be13d866855c5fcd655

src/pkg/caendr/caendr/services/sql/dataset/_templates.py
vlagrassa: Fixed in 3473d8e0c0d73e0daf16cfa3723fbf57223bf9a8

src/pkg/caendr/caendr/services/indel_primer.py
vlagrassa: Yes, I believe so.  They're only used in this file to set the min/max size that an indel can be to be displayed.

Where would we store them in the database?

src/pkg/caendr/caendr/services/sql/dataset/_env.py
vlagrassa: Started working on better guards across the codebase in new branch: https://github.com/AndersenLab/CAENDR/tree/enhancement/get-env-vars

src/modules/site/base/views/mapping.py
r-vieira: thank you

src/modules/site/base/views/mapping.py
r-vieira: Excellent!

src/modules/site/base/views/mapping.py
r-vieira: thanks for the nuanced response! My original comment was rendered incorrectly, it was missing the `<id>` bit, sry about that if made the comment unclear. The main concern was about `mapping` being `None`, which has been covered. 

Regarding your question about the failed calculation -- if the pipeline errors out, there are two levels of logs that we capture -- the "platform error" which happens before the nextflow pipelines start, and then the errors generated by nextflow. We can look at how these are displayed in other tasks and unify them across tools. Let's discuss.

src/pkg/caendr/caendr/models/datastore/wormbase.py
r-vieira: The unity test should be a different PR, no need to submit it for this release version.

src/pkg/caendr/caendr/services/indel_primer.py
r-vieira: Good question -- most likely DataStore with a new `kind=settings`, and a single row for the tool. Also, our code needs to create this row with default values (seed values included in the repo). End-users will then be able to change them via UI option in the Tool/Versions/Settings.

r-vieira: this also should be a new PR and not included in this release.

src/modules/site/base/views/mapping.py
vlagrassa: (Submission being `None`, as well as some of my concerns above, fixed in 0f531415cf290a2cc0ed24dd25e74b47b132c058)

src/modules/site-v2/base/views/tools/genetic_mapping.py
vlagrassa: Fixed in 0f531415cf290a2cc0ed24dd25e74b47b132c058 -- see also discussion at https://github.com/AndersenLab/CAENDR/pull/145#discussion_r1132691789

src/pkg/caendr/caendr/models/datastore/wormbase.py
vlagrassa: I believe so, in normal usage: other than this line, `self._value` can only be set on lines 14 or 32, where it's set to `None` or `value.upper()` respectively.  This means it should be impossible to get a `WormbaseVersion` object with a lowercase string in `self._value`, unless the `_value` field is modified directly in an external source.  I can add a check for this just in case.

vlagrassa: Fixed in ac3c502ee2c29f8c394fed6d9977369d2c33405f

src/modules/site-v2/base/views/tools/genome_browser.py
vlagrassa: This value isn't actually needed; removed in 1d2373488ff9e1a7c1a9f576eda55acc48012c2d.

src/modules/site-v2/base/views/tools/genetic_mapping.py
vlagrassa: Fixed in b1bae04850b89f5eb3dee32dc9fa474feb4b4698

src/modules/site-v2/base/views/tools/genetic_mapping.py
vlagrassa: Ad-hoc fix in 415835038d131d34aeded420e44f3ca7b8dcb36b.  Will update to a better version (using Flask Form validation) soon.

src/modules/site-v2/base/views/about/about.py
vlagrassa: The two static `"c_elegans"` references in this file are actually unnecessary -- removed in 1df7d83b8c2c915cccfd8fd9ad1b296727b39c5b.

A few more static references remain in the codebase, e.g. in the Downloads workflow.  Since we'll have to overhaul this page anyways, I'm holding off on it for now.

src/pkg/caendr/caendr/services/user.py
```diff
@@ -7,15 +7,17 @@
   'admin': 'Admin'
 }
 
+# TODO: Replace with Entity.query_ds call? Should this return as User Entity class?
```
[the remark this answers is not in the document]
vlagrassa: It looks like there are two places where this function is used:

- On the registration page, to check if this email is already associated with an account.  Since all we care about here is whether the email is in use or not, I think it's alright to query by email.
  ```
  def validate_email(form, field):
    existing_user = get_local_user_by_email(field.data)
    if len(existing_user):
      raise ValidationError("Email already exists")
  ```
  Code [here](https://github.com/AndersenLab/CAENDR/blob/feature/multiple-species-etl-ops/src/modules/site-v2/base/forms/forms.py#L113).

- To reset a user's password.  I'm less sure what to do here -- the existing code just takes the first result, and uses that username to generate a one time token.
  ```
  email = request.form.get("email")
  users = get_local_user_by_email(email)

  if not users or not users[0]:
    logger.error(f"Failed to retrieve LOCAL user for email: {email}")
    flash('Unable to reset your credentials at this time. Please try again later.', 'error')
    return redirect('/')
    
  user = users[0]
  ```
  Code [here](https://github.com/AndersenLab/CAENDR/blob/feature/multiple-species-etl-ops/src/modules/site-v2/base/views/user.py#L60).

How should we handle the password reset page?

src/modules/site/base/views/about.py
vlagrassa: Removed in 1df7d83b8c2c915cccfd8fd9ad1b296727b39c5b.  See discussion at https://github.com/AndersenLab/CAENDR/pull/145#discussion_r1132918582.

r-vieira (approved):
Merged!

PR #155, opened 2023-03-07
added the map to homepage, exposed data on render side

orzut:


src/modules/site-v2/templates/primary/map.html
```diff
@@ -0,0 +1,252 @@
+
+<div>
+        <div id="map" style="height: 500px; width:100%"></div>
```
r-vieira, line 3: @orzut: we prefer moving the styles out of the code whenever possible. In this case, consider moving the height&width declarations to the existing <style/> section on the file, or if there is a references `.css` style file specific to this page, you can move this declaration there as well.

src/modules/site-v2/templates/primary/map.html
```diff
@@ -0,0 +1,252 @@
...
+        <div class='text-center'>
+          <small><strong>Hover over or click a pin to see information about a <i>C. elegans</i> wild isolate</strong></small>
+        </div>{# /text-center #}
+
+      <!-- <div class="col-md-4">
```
r-vieira, line 8: @orzut: we can remove this unused code.

src/modules/site-v2/templates/primary/map.html
```diff
@@ -0,0 +1,252 @@
...
+     </p>
+    </div>{# /col-md-4 #} -->
+
+</div>
+
+
+
+{% block script %}
+
+<script>
```
r-vieira, line 58: @orzut: whenever possible, we want our JS to be inside of `<script type="module">` tags. This ensures that each JS block is isolated from other blocks on the same page. See if adding the attribute `type="module"` breaks anything and consider adding it.

src/modules/site-v2/templates/primary/map.html
```diff
@@ -0,0 +1,252 @@
...
+
+// Icons
+{% autoescape off %}
+var icon_norm = L.icon({
```
r-vieira, line 62: @orzut: i know we are porting [older] existing code, but wanted to keep this in the radar: for new code, we prefer using `let` and `const` and avoid using `var`.

r-vieira (changes requested):
Thanks, @orzut! Great PR! Please see inline comments and update with PR with the suggestions.

r-vieira (approved):
Looks good!

PR #167, opened 2023-04-04
Bugfix: Remove '$' from environment variables

vlagrassa:
Should fix two Indel Primer issues on dev site:
- Strain lists don't populate when species selected
- Jobs fail because incorrect filename is used


r-vieira (approved):
looks good.

PR #171, opened 2023-04-06
Standardize loading & retrieving environment variables

vlagrassa:
- Get vars using `get_env_var`, which checks that environment and variable are both defined, as well as string Template conversions
- Load environment with `load_env`, which tracks whether env is already loaded and which throws a unique error if load fails
- Move utility functions to `utils/env.py`, for interpreting & logging environment variables


vlagrassa:
(I've tested the new Indel Finder module locally, and can confirm the new code works as expected)

r-vieira (approved):
Looks good.

PR #172, opened 2023-04-11
Added layout for People page

natalieroman:


src/modules/site-v2/templates/about/people.html
```diff
@@ -1,83 +1,123 @@
...
-
+        <div class="col col-md-9 mt-5">
+            <!-- Staff section -->
+            <div class="row">
+                <h2 id="staff">Staff</h2>
+            </div>
+            <!-- /Row -->
+            <div class="row row-cols-1 row-cols-md-3 g-5 mb-5">
+                <!-- Card Column -->
+                {% for p in profiles['staff'] %}
```
r-vieira, line 36: @vlagrassa - instead of `profiles['staff']` we should use a more defensive approach and have something like `profiles.get('staff', [])`. This will not break if the key is missing.

src/modules/site-v2/templates/about/people.html
```diff
@@ -1,83 +1,123 @@
...
+            <!-- /Row -->
+            <!-- /Staff section -->
+            <!-- Scientific Advisory Committee section -->
+            <div class="row">
+                <h2 class="mt-5" id="SAC">Scientific Advisory Committee</h2>
+            </div>
+            <!-- /Row -->
+            <div class="row row-cols-1 row-cols-md-3 g-5 mb-5">
+                <!-- Card Column -->
+                {% for p in profiles['committee']|sort(attribute='last_name') %}
```
r-vieira, line 73: @vlagrassa same here (use `profiles.get(..., [])`)

src/modules/site-v2/templates/about/people.html
```diff
@@ -1,83 +1,123 @@
...
+            <!-- /Row -->
+            <!-- /Staff section -->
+            <!-- Collaborators section -->
+            <div class="row">
+                <h2 class="mt-5" id="collaborators">Collaborators</h2>
+            </div>
+            <!-- /Row -->
+            <div class="row row-cols-1 row-cols-md-3 g-5 mb-5">
+                <!-- Card Column -->
+                {% for p in profiles['collaborators']|sort(attribute='last_name') %}
```
r-vieira, line 99: @vlagrassa  here too.

r-vieira (approved):
looks good!

vlagrassa:
Fixed lookups in new branch - see https://github.com/AndersenLab/CAENDR/pull/177

PR #186, opened 2023-04-24
Standardize Job Submission Flow

vlagrassa:
- When each tool sends a POST request to create a new job, it receives a standardized JSON object as a response
- All jobs trigger the confirmation modal after a successful submission
- Heritability & Mapping display an error message for malformed input data or server error
  - Still TODO: Handle server error response in Indel Finder
- All tools open cached report if found
  - Indel Finder opens in new tab
  - Heritability & Mapping open in current tab
- Cleaned up view functions
  - Functions no longer pass locals
  - Report unpacking logic moved to `services` folder files
- Standardized endpoint names & URLs
- Update report list tables
  - Merged all six pages (my reports / all reports x all three tools) into one HTML template
  - Updated table styling to match designs on BaseCamp


src/pkg/caendr/caendr/services/sql/etl/strains.py
```diff
@@ -6,16 +6,18 @@
 from caendr.services.cloud.secret import get_secret
 from caendr.models.sql import Strain
 
-ANDERSEN_LAB_STRAIN_SHEET = [
-  get_secret('ANDERSEN_LAB_STRAIN_SHEET'),
+ANDERSEN_LAB_STRAIN_SHEETS = [
+  get_secret('ANDERSEN_LAB_STRAIN_SHEET'),            # elegans
```
r-vieira, line 10: @vlagrassa can we rename this secret to include the `_C_ELEGANS` suffix?

r-vieira (approved):
looks great!

PR #190, opened 2023-04-25
Feature/new design isotype db

orzut:


src/modules/site-v2/base/static/img/how to move button right in bootstrap - Google Search.html
r-vieira, line 1: @orzut -- check if this file is needed.

src/modules/site-v2/base/views/auth/oauth.py
```diff
@@ -56,6 +60,29 @@ def authorized(google_bp, token):
...
-  return assign_access_refresh_tokens(user.name, user.roles, session.get("login_referrer"))
+  return new_resp
+
+
+def transfer_cart(resp, user):
+  # checks if a user has a local cart
+  cartID = request.cookies.get('cartID')
+  if cartID:
+    local_cart = Cart(cartID)
+    if len(local_cart['items']):
```
r-vieira: @orzut this is a minor stylistic note, but we prefer code-guards that return early (and keep indentation minimal). 
For instance, when there are zero items, immediately delete the cookie and return. That way the rest of the flow can remain at the function level. Example:
``` 
if len(local_cart['items']) == 0:
  res.delete_cookie('cartID')
  return 

# continue code at this level of indentation.
users_cart = Cart.lookup_by_user(user['email'])
# (...)
```

src/modules/site-v2/base/views/auth/oauth.py
```diff
@@ -56,6 +60,29 @@ def authorized(google_bp, token):
...
+  resp = make_response(assign_access_refresh_tokens(user.name, user.roles, session.get("login_referrer")))
+  new_resp = transfer_cart(resp, user)
   flash("Successfully logged in!", 'success')
-  return assign_access_refresh_tokens(user.name, user.roles, session.get("login_referrer"))
+  return new_resp
+
+
+def transfer_cart(resp, user):
+  # checks if a user has a local cart
+  cartID = request.cookies.get('cartID')
```
r-vieira: @orzut -- one more stylistic note: in python the norm is to use lower-case var names with underscores, instead of `cartID =` consider using `cart_id =`. 

The cookie name is fine as is, and we might want to promote it to an environment variable (env/ENV/global.env) that is unique to each environment. For instance, on the env/development/global.env it could be: `MODULE_SITE_CART_COOKIE_NAME=caendr-dev-cart-id` and then use it here as:
```
MODULE_SITE_CART_COOKIE_NAME = os.getenv('MODULE_SITE_CART_COOKIE_NAME')
cart_id = request.cookies.get(MODULE_SITE_CART_COOKIE_NAME)
```

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +146,144 @@ def strains_catalog():
...
+                  "to": [order_obj["email"]],
+                  "cc": config.get("CC_EMAILS"),
+                  "subject": "CeNDR Order #" + str(order_obj["invoice_hash"]),
+                  "text": ORDER_SUBMISSION_EMAIL_TEMPLATE.format(**order_obj)})
+
+      # Save to google sheet
+      add_to_order_ws(order_obj)
+
+      # Mark the cart is deleted
+      users_cart.delete_cart()
```
r-vieira: @orzut  -- since we are doing a 'soft-delete', we might want to rename this method to #soft_delete(). Also, since the method applies to `Cart`, we can remove the suffix `_cart` from the method name.

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +146,144 @@ def strains_catalog():
...
+      order_obj['invoice_hash'] = str(uuid.uuid4()).split("-")[0]
+      order_obj["order_confirmation_link"] = url_for('request_strains.order_confirmation', invoice_hash=order_obj['invoice_hash'], _external=True)
+      send_email({"from": "no-reply@elegansvariation.org",
+                  "to": [order_obj["email"]],
+                  "cc": config.get("CC_EMAILS"),
+                  "subject": "CeNDR Order #" + str(order_obj["invoice_hash"]),
+                  "text": ORDER_SUBMISSION_EMAIL_TEMPLATE.format(**order_obj)})
+
+      # Save to google sheet
+      add_to_order_ws(order_obj)
```
r-vieira, line 234: @orzut -- it might be helpful to notify site-admins when the application is unable to save the order to the google sheets. Does the `add_to_order_ws()` function throw an exception we could catch here?

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +146,144 @@ def strains_catalog():
...
+
+      # Mark the cart is deleted
+      users_cart.delete_cart()
+      users_cart.save()
+
+      # flash("Thank you for submitting your order! Please follow the instructions below to complete your order.", 'success')
+      flash("Thank you for submitting your order! Please follow the instructions below to complete your order.", 'success')
+
+      # delete cartID from cookies
+      if cartID:
```
r-vieira: prefer the early return, eg :
```
if cartID is None:
        return redirect(url_for("request_strains.order_confirmation", invoice_hash=order_obj['invoice_hash']), code=302)

# continue at this indentation
resp = ...
```

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +146,144 @@ def strains_catalog():
...
+    
+    return resp
+
+@strains_bp.route('/checkout', methods=['PUT'])
+@jwt_required(optional=True)
+def order_page_remove():
+  """ This view handles removing items from the cart """
+  form = OrderForm()
+  user = get_current_user()
+  item_to_remove = request.get_json()['itemToRemove']
```
r-vieira: be careful when doing key look-ups in python dicts, if a key doesn't exist, it will throw a `KeyError`. An alternative that is usually recommended is to use `dict.get('itemToRemove', 'defaultValueIfKeyDoesntExist')`. By default when using .get('key') and not passing a default value, if the key doesn't exist, it will return `None` (which is safer) than a `KeyError`

src/modules/site-v2/base/views/strains.py
```diff
@@ -205,7 +293,23 @@ def order_confirmation(invoice_hash):
...
+@cache.memoize(60*60)
+def strains_submission_page():
+  """
+      Google form for submitting strains
+  """
+  title = "Strain Submission"
+  # TODO: Move this to configurable location
+  #STRAIN_SUBMISSION_FORM = '1w0VjB3jvAZmQlDbxoTx_SKkRo2uJ6TcjjX-emaQnHlQ'
+  #strain_submission_url = f'https://docs.google.com/forms/d/{STRAIN_SUBMISSION_FORM}/viewform?embedded=true'
+  strain_submission_url = "https://docs.google.com/forms/d/1w0VjB3jvAZmQlDbxoTx_SKkRo2uJ6TcjjX-emaQnHlQ/viewform?embedded=true"
```
r-vieira: Since this URL will change based on what environment the application is running, it should be promoted to a env/development/global.env variable. (and it should be propagated to other environments, eg: main).

r-vieira (changes requested):
Looks great!! Please review the inline comments and adjust as needed.

src/modules/site-v2/base/views/strains.py
orzut: No, none of the google sheets functions don't throw any exceptions

src/modules/site-v2/templates/errors/400.html
```diff
@@ -5,5 +5,5 @@
 {% block content %}
   <h1>Internal Server Error :(</h1>
   <h3>We're having some problems</h3>
-  <h3>Instead, look at all these <a href="{{ url_for('request_strains.strains') }}">worms</a>!</h3>
+  <h3>Instead, look at all these <a href="{{ url_for('request_strains.request_strains') }}">worms</a>!</h3>
```
natalieroman: I discussed with Erik and he said it makes sense to link to the homepage on the error pages. I updated [here](https://github.com/AndersenLab/CAENDR/commit/d3c29c5cec04dac0ece8fbe62bbcfa1345bbe8ba).

src/modules/site-v2/templates/errors/404.html
```diff
@@ -5,5 +5,5 @@
 {% block content %}
   <h1>Page Not Found :(</h1>
   <h3>What you were looking for is just not here.</h3>
-  <h3>Instead, look at all these <a href="{{ url_for('request_strains.strains') }}">worms</a>!</h3>
+  <h3>Instead, look at all these <a href="{{ url_for('request_strains.request_strains') }}">worms</a>!</h3>
```
natalieroman: [Same here.](https://github.com/AndersenLab/CAENDR/pull/190/files/edc27cfd2afdc49660981c8b66115d1a3ebc0654#r1179603318)

env/development/global.env
```diff
@@ -140,3 +141,6 @@ JWT_ACCESS_CSRF_FIELD_NAME=jwt_csrf_token
 
 ETL_LOGS_BUCKET_NAME=caendr-dev-etl-logs-4objm
 SENTRY_URL="https://78a94a7e99a546679ae5d368f3775eb2@o353488.ingest.sentry.io/6166204"
+
+MODULE_SITE_CART_COOKIE_NAME="caendr-dev-cart-id"
```
vlagrassa: Just make sure an equivalent variable exists in `env/main/global.env`

src/modules/site-v2/base/views/auth/oauth.py
```diff
@@ -1,18 +1,26 @@
...
 from flask_dance.consumer import oauth_authorized
 
 from base.utils.auth import assign_access_refresh_tokens
 
 from caendr.models.datastore import User
+from caendr.models.datastore.cart import Cart
 from caendr.utils.data import unique_id
 from caendr.services.cloud.secret import get_secret
 
+MODULE_SITE_CART_COOKIE_NAME = os.getenv('MODULE_SITE_CART_COOKIE_NAME')
```
vlagrassa: You can use:
```
from caendr.utils.env import get_env_var
...
MODULE_SITE_CART_COOKIE_NAME = get_env_var('MODULE_SITE_CART_COOKIE_NAME')
```
...to get the environment variable a bit more safely.  I'll add documentation for this to the README.

src/modules/site-v2/base/views/strains.py
```diff
@@ -34,6 +38,8 @@
 from caendr.services.email import send_email, ORDER_SUBMISSION_EMAIL_TEMPLATE
 from caendr.services.cloud.sheets import add_to_order_ws, lookup_order
 
+MODULE_SITE_CART_COOKIE_NAME = os.getenv('MODULE_SITE_CART_COOKIE_NAME')
```
vlagrassa: Same deal here - `get_env_var`

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +149,146 @@ def strains_catalog():
...
+  cart_id = request.cookies.get(MODULE_SITE_CART_COOKIE_NAME)
+  if user:
+    users_cart = Cart.lookup_by_user(user['email'])
+  elif cart_id:
+    users_cart = Cart(cart_id)
+  else:
+    users_cart = Cart(**{'items': []})
+
+  if 'order_form' in request.form:
+    if form.validate_on_submit():
```
vlagrassa: Might be helpful to turn these nested `if`-statements into a flat sequence of guards, along the lines of:
```
if 'order_form' not in request.form:
  # return error message

if not form.validate_on_submit():
  # return error message

# continue with the code
```

This isn't strictly necessary, but it makes it more obvious what conditions you're checking and how you handle error cases.  Happy to give advice on what errors to raise, if you'd like!

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +149,146 @@ def strains_catalog():
...
+
+      # Mark the cart is deleted
+      users_cart.soft_delete()
+      users_cart.save()
+
+      # flash("Thank you for submitting your order! Please follow the instructions below to complete your order.", 'success')
+      flash("Thank you for submitting your order! Please follow the instructions below to complete your order.", 'success')
+
+      # delete cartID from cookies
+      if cart_id is None:
```
vlagrassa: This is nitpicky, so feel free to ignore, but a clearer way to structure this might be:
```
resp = make_response(redirect(url_for("request_strains.order_confirmation", invoice_hash=order_obj['invoice_hash']), code=302))

if cart_id is not None:
  resp.delete_cookie(MODULE_SITE_CART_COOKIE_NAME)

return resp
```

This makes it more obvious that you're redirecting to the same place in either case.

(Please double-check that I copied everything correctly, of course! Also, I'm pretty sure you can return the `make_response` version even if you're not deleting a cookie, but please double-check that if you go in this direction.)

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +149,146 @@ def strains_catalog():
...
+      flash("You must select strains/sets from the catalog", 'error')
+      return redirect(url_for("request_strains.strains_catalog"))
+ 
+    for item in added_items:
+      users_cart.add_item(item)
+    users_cart.save()
+
+    resp = make_response(jsonify({'status': 'OK'}))
+    if not user:
+      resp.set_cookie(MODULE_SITE_CART_COOKIE_NAME, users_cart.name, max_age=60*60*24*14)
```
vlagrassa: Since you're working with env vars anyways, might make sense to move the max cart age to the `.env` files.  You'd have to look into how to store/read an int value, since I think they're all strings by default.

If it makes sense to add a `var_type` parameter to the `get_env_var` function, so that you can do something like below, you're more than welcome to do so:
```
COOKIE_AGE = get_env_var('MODULE_SITE_CART_COOKIE_AGE', var_type=int)
```
(I would call it `type` but I believe that's a keyword in Python.)

vlagrassa: This one's totally optional -- don't stress about it if you have other work to do!

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +149,146 @@ def strains_catalog():
...
+@strains_bp.route('/checkout', methods=['PUT'])
+@jwt_required(optional=True)
+def order_page_remove():
+  """ This view handles removing items from the cart """
+  form = OrderForm()
+  user = get_current_user()
+  item_to_remove = request.get_json()['itemToRemove']
+
+  # get cart
+  if user:
```
vlagrassa: Is it possible for either or both of these conditions to fail? I.e. what if the user has no cart on GCP and no cookie?

I can imagine a situation where this gets a `None` value for the cookie, creates/gets a Cart with an ID `None`, and operates on that.  Might not technically break anything, but not exactly desirable either.

(My guess is the front-end site is set up so this would never realistically happen, but on the back end, you don't want to rely on that.)

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +149,146 @@ def strains_catalog():
...
+
+  return jsonify({'status': 'OK'}), 200
+
+@strains_bp.route('/checkout')
+@jwt_required(optional=True)
+def order_page_index():
+  """ This view handles the checkout page """
+  form = OrderForm()
+  user = get_current_user()
+  if user:
```
vlagrassa: Similar question here - what if neither of these branches would find a valid cart?

Using the example above -- if `request.cookies.get` returns `None`, and if the site uses that as an ID, you might end up with multiple non-logged-in users accessing a single "default" `Cart` object.

src/modules/site-v2/base/views/strains.py
```diff
@@ -143,59 +149,146 @@ def strains_catalog():
...
+    users_cart = Cart.lookup_by_user(user['email'])  
+  else:
+    cart_id = request.cookies.get(MODULE_SITE_CART_COOKIE_NAME)
+    users_cart = Cart(cart_id)
 
+  cartItems = users_cart['items'] 
+  for item in cartItems:
+    item_price = users_cart.get_price(item)
```
vlagrassa: Quick semantic point - `get_price` should probably be a `@classmethod` of `Cart`, since its value shouldn't change based on the particular cart you'd talking about.  Then you'd probably want to use it as:
```
  item['price'] = Cart.get_price(item)
```

src/modules/site-v2/templates/order/order_confirm.html
```diff
@@ -96,49 +65,52 @@ <h4 class="list-group-item-heading">Address</h4>
...
                   <td class="price text-right">$65</td>
                 </tr>
                 {% endif %}
-
+  
                 <tr>
                 <td><strong>Total</strong></td>
                 <td></td>
-                <td class="price text-right"><strong>{{ order_obj["total"] }}</strong></td>
+                <td class="price text-right"><strong>${{ order_obj["total"] }}</strong></td>
```
vlagrassa: Just be careful about using `[]` lookup notation directly -- usually `.get()` is safer.  Lets you build in a default error value.

src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -0,0 +1,93 @@
...
+    if not self['items']:
+      self['items'] = []
+  
+  ## Properties List ##
+
+  @classmethod
+  def get_props_set(cls):
+    return {
+      *super().get_props_set(),
+      'user',
```
vlagrassa: Since last time we talked about this, I ended up making a `UserOwnedEntity` class to handle User ownership over pipeline jobs.  Might be worth leveraging that here.

I think all you'd have to do is change the declaration to `class Cart(UserOwnedEntity)` and remove `user` here (since it'll be handled by the parent class).  Happy to discuss further, if you'd like!

src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+    # User doesn't have a cart
+    except NotFoundError:
+      return Cart(**{'user': email, 'items': []})
+
+    # Multiple carts found for user
+    # TODO: get all carts that match the user, sort by date created on and return the latest one?
+    except NonUniqueEntity:
+      return 'multiple carts found'
```
vlagrassa: You'll probably want this to return either a `Cart` or `None`, just for consistency.

If you want to return one of the carts (e.g. the most recent one), you could do:
```
except NonUniqueEntity as ex:
  # choose the right Cart from ex.matches
```
You could also just let the error propagate for now, and come back to it later...

src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -0,0 +1,93 @@
...
+        return
+      else:
+        continue  
+
+
+  def soft_delete(self):
+    self['is_deleted'] = True
+
+
+  def get_price(self, item):
```
vlagrassa: Same note as above -- this should probably be a class method.

src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -0,0 +1,93 @@
...
+        return
+    self['items'].append(item)
+
+
+  def remove_item(self, item):
+    for cartItem in self['items']:
+      if cartItem['name'] == item:
+        self['items'].remove(cartItem)
+        return
+      else:
```
vlagrassa: A bit pedantic, but technically you don't need this `else` clause

src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -0,0 +1,93 @@
...
+      if cartItem['name'] == item['name']:
+        return
+    self['items'].append(item)
+
+
+  def remove_item(self, item):
+    for cartItem in self['items']:
+      if cartItem['name'] == item:
+        self['items'].remove(cartItem)
+        return
```
vlagrassa: Genuine question - is it possible for the same item to end up in the cart multiple times? And if so, should this function keep going, to make sure it removes all copies?

src/pkg/caendr/caendr/models/datastore/cart.py
vlagrassa: Completely optional extension to this -- you could actually move this to the `PRICES` class, if you wanted to.  I'd recommend something like:
```
class PRICE:
  
  # insert price variables

  def lookup(item):
    # return the price
```

Then you could use it as:
```
  item['price'] = PRICE.lookup(item)
```

This is totally up to you.  I personally think it's cleaner, since it bundles the pricing logic together in one place, but it's not like it fixes a potential bug or anything.

src/pkg/caendr/caendr/models/datastore/entity.py
```diff
@@ -388,6 +388,45 @@ def query_ds_unique(cls, key, val, required=False, safe=False):
...
 
 
+  @classmethod
+  def query_ds_not_deleted(cls, key, val, required=False):
+
+    # Run query with given key and val
+
+    matches = cls.query_ds(filters=[(key, '=', val)])
+
+    # TODO: Filter out any elements in "matches" that are deleted
```
vlagrassa: Very nitpicky point, but might want to clean up the spacing here & remove this `TODO` message.  I know we're pretty lax about `TODO`s, but since you do actually implement what it describes in this PR, I'd recommend cleaning it up.

vlagrassa (changes requested):
Looking very good!  I have a few recommendations for code clarity, and one or two potential sources of bugs I think it would be good to patch up.

The major takeaway, in my opinion, is: on the back-end, you don't want to rely on the front-end always giving you good data.  It's better to explicitly check for failure cases and return a descriptive error, so we know what happened if it goes wrong later down the line.

src/modules/site-v2/templates/_includes/isotype-database.html
```diff
@@ -1,749 +1,80 @@
...
+  }
+  thead {
+    background-color: #e1eae4;
+  }
+
+</style>
+
+{% endblock %}
+
+<div class="table-responsive my-5 isotypeDatabase">
```
natalieroman: I'd recommend moving the `table-responsive` div to the home.html template since that just feels a little bit safer and less likely to break the design in the future.

I also think you can remove the `isotypeDatabase` class, because all that's doing is limiting the height. I think there's enough room in the design to show the full height.

src/modules/site-v2/templates/_includes/isotype-database.html
```diff
@@ -1,749 +1,80 @@
...
+    <caption class="visually-hidden">A list of strains sortable by species that includes various data about each
+      individual strain.</caption>
+    <thead class="table-secondary align-middle">
+      <tr>
+        <th scope="col">#</th>
+        <th scope="col">Species</th>
+        <th scope="col">Strain</th>
+        <th scope="col">Isotype</th>
+        <th scope="col">Previous Names</th>
+        <th scope="col">Release</th>
```
natalieroman: Need to add a column for "Source Lab" between "Previous Names" and "Release." Erik asked to add this column during the design review.

natalieroman (changes requested):
The isotype database looks awesome! Just a couple of minor changes.

src/pkg/caendr/caendr/models/datastore/cart.py
vlagrassa: (This branch is not merged yet, so deferring this issue for later. Discussed w @orzut in person.)

vlagrassa (approved):
Looks good to me!

natalieroman (approved):
Looks good to me!

env/development/global.env
```diff
@@ -44,6 +44,10 @@ MODULE_SITE_BUCKET_PUBLIC_NAME=caendr-dev-site-public
 MODULE_SITE_BUCKET_PRIVATE_NAME=caendr-dev-site-private
 
 MODULE_SITE_SENTRY_NAME="caendr-dev-site-sentry-dsn"
+MODULE_SITE_CART_COOKIE_NAME="caendr-dev-cart-id"
+
+#14 days 
+MODULE_SITE_CART_COOKIE_AGE=1209600
```
r-vieira: @orzut good! Wondering if we could add the `unit` of time to the name of the var? e.g. use `MODULE_SITE_CART_COOKIE_AGE_SECONDS` instead of `MODULE_SITE_CART_COOKIE_AGE`

r-vieira (approved):
looks good! Please submit a new PR to add the unit of time to the cookie name.

r-vieira (approved):
looks good! Please submit a new PR to add the unit of time to the cookie name.

PR #195, opened 2023-05-10
Add C. tropicalis

vlagrassa:
Minor update -- most changes are on GCP.

### DIFF
--- a/src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
+++ b/src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
@@ -58,6 +58,12 @@ def results_columns():
     },
   ]
 
+def try_get_sv_strains(species):
+  try:
+    return get_sv_strains(species)
+  except:
+    return []
+
 
 ## Data Endpoints
 
@@ -88,7 +94,7 @@ def get_tracks():
 @jwt_required()
 def get_strains():
   return jsonify({
-    species: get_sv_strains( species ) for species in SPECIES_LIST.keys()
+    species: try_get_sv_strains( species ) for species in SPECIES_LIST.keys()
   })
 
 
--- a/src/pkg/caendr/caendr/services/sql/etl/strains.py
+++ b/src/pkg/caendr/caendr/services/sql/etl/strains.py
@@ -7,15 +7,15 @@ from caendr.services.cloud.secret import get_secret
 from caendr.models.sql import Strain
 
 ANDERSEN_LAB_STRAIN_SHEETS = [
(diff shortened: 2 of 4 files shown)


src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
```diff
@@ -58,6 +58,12 @@ def results_columns():
     },
   ]
 
+def try_get_sv_strains(species):
+  try:
+    return get_sv_strains(species)
+  except:
+    return []
```
r-vieira, line 65: @vlagrassa -- we prefer avoiding silent errors. If appropriate, consider logging here that the species was not found. We can always use that data-point in the logs to troubleshoot issues.

src/pkg/caendr/caendr/services/sql/etl/strains.py
```diff
@@ -4,18 +4,21 @@
...
 from caendr.models.sql import Strain
 
+# Get list of Google Sheet IDs for each species
+# Expects secret names with prefix "ANDERSEN_LAB_STRAIN_SHEET_"
+# and species ID in all caps
 ANDERSEN_LAB_STRAIN_SHEETS = [
-  get_secret('ANDERSEN_LAB_STRAIN_SHEET'),            # elegans
-  get_secret('ANDERSEN_LAB_STRAIN_SHEET_C_BRIGGSAE'), # briggsae
-  # tropicalis
+  get_secret(f'ANDERSEN_LAB_STRAIN_SHEET_{species_name.upper()}')
```
r-vieira, line 14: @vlagrassa we should have a list of 'allowed' species names that can be read from the ENV (i.e. validate the species name before looking it up).

r-vieira (changes requested):
Looks good -- see note about validating the species name before looking in ENV

src/pkg/caendr/caendr/services/sql/etl/strains.py
vlagrassa: Normally I'd check that by making sure the `species_name` is in `SPECIES_LIST`, but here I'm generating it from the species list itself.  Should I check to make sure the env var exists before trying to get it, maybe?

Alternatively, I could go back to hard-coding the secret names.  It might not make sense to generate them dynamically, now that I think about it.

src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
vlagrassa: Fixed in 6c8c326d22b50ff43e749d8a3519d88249b796ae

PR #196, opened 2023-05-10
Fix heritability reports

vlagrassa:
- Minor updates to reports page so it still displays even with missing data
- Revert to accepting any trait name as second column header (in input data)


src/modules/site-v2/base/views/tools/heritability_calculator.py
```diff
@@ -239,7 +239,7 @@ def report(id):
     ready = True
 
   # If report hasn't finished computing yet, display a waiting page
-  except UnfinishedReportError:
+  except UnfinishedReportError as ex:
     data, result = ex.data, None
```
r-vieira: @vlagrassa is there a less convoluted way (non-exception-based) way of passing data to the frontend in this code path?

r-vieira (approved):
Looks good -- see inline comment about populating the data from the exception.

src/modules/site-v2/base/views/tools/heritability_calculator.py
vlagrassa: Fixed in e67d52b0afcffaa16bb95778160f86fb855e7fde.  Waiting on PR to figure out empty Indel Finder results.

r-vieira (approved):
looks good!

PR #197, opened 2023-05-10
Manage user ownership with UserOwnedEntity class

vlagrassa:
Moves logic for Entity with user field to a new superclass, `UserOwnedEntity`.  This class looks up name & email based on the associated user ID, and briefly caches User objects.

NOTE: I've set the cache age for 1 hour.  The primary goal of the cache is to handle the report lists with a single lookup for each unique user -- otherwise, every job would look up the same few user objects in rapid succession.


src/modules/site-v2/templates/admin/profile/list.html
```diff
@@ -26,7 +26,7 @@
             <td> {{ profile['first_name'] }} </td>
             <td> {{ profile['org'] }} </td>
             <td> {{ profile['email'] }} </td>
-            <td> {{ profile['prof_roles'] }} </td>
+            <td> {% for role in profile['prof_roles'] %}{{ role.code }}{% if loop.revindex0 > 0 %}, {% endif %}{% endfor %} </td>
```
r-vieira, line 29: props on using `loop.revindex0`!

r-vieira (approved):
Looks good!

PR #201, opened 2023-05-10
Feature/new design fullscreen

natalieroman:
Added the "view fullscreen" functionality to tools


src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
```diff
@@ -443,8 +452,32 @@ <h1 class="modal-title fs-5" id="infoModalLabel">Track Information</h1>
   return genotype_set.filter( gt => gt['GT'] == genotype_val ).map( gt_label ).join(" ");
 }
 
+// Open and Close Fullscreen Buttons
+var fullscreenBrowser = document.getElementById("browser");
```
r-vieira, line 456: Note: use `let` or `const` instead of `var` when writing new code.

r-vieira (approved):
approved!

PR #206, opened 2023-05-16
Remove obsolete Gene Browser Tracks module

vlagrassa:
Old gene_browser_tracks module is deprecated and will be removed.

Separately, admin page needs a new Gene Browser Tracks page to manage the list of tracks. Updated existing endpoint to display the existing template with no data, so we can fill in the new page needs later.


r-vieira (approved):
looks good!

PR #208, opened 2023-05-19
Disable new species in Heritability dropdown

vlagrassa:


src/modules/site-v2/base/forms/forms.py
```diff
@@ -87,8 +88,11 @@ class FileUploadForm(FlaskForm):
   label = StringField('Description:', validators=[Required(message='You must include a description of your data.')])
   file = FileField('Select file:', render_kw={'accept': '.tsv'})
 
-class HeritabilityForm(FileUploadForm):
-  pass
+# class HeritabilityForm(FileUploadForm):
+class HeritabilityForm(FlaskForm):
+  species = SpeciesSelectField(exclude_species=['c_briggsae', 'c_tropicalis'])
```
r-vieira, line 93: minor enhancement -- we should use class enums for these string literals whenever possible. (eg: Species.BRIGGSAE)

r-vieira (approved):
Looks good

PR #213, opened 2023-05-22
Fix/alerts

natalieroman:
Added a close button to alerts so the user can dismiss them. Added alerts about orders not shipping until fall 2023 to the "Request Strains" landing page and the "Strain Catalog" page. Re-did "request strains" template using bootstrap because custom CSS was causing the alert to look strange.


r-vieira (approved):
looks good!

PR #215, opened 2023-05-23
Update SVA Postgres table pipeline

vlagrassa:
Also stops the Indel Finder IGV Browser from throwing an error if the Divergent Regions track file can't be found.  This doesn't break the tool, and we don't have it yet for C. tropicalis.


src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -24,7 +24,15 @@ function toggle_input(idOrEl, val) {
 
 
 function fetch_json(url) {
-  return fetch(url).then(res => res.json())
+  return new Promise((resolve, reject) => {
+    fetch(url).then(res => {
+      if (res.status == 404) {
```
r-vieira, line 29: the inverted logic might be safer -- only resolve when we have status===200. (Otherwise, server can send 5XX and other 4XX) codes.

src/modules/site-v2/templates/admin/etl/list.html
```diff
@@ -86,7 +86,11 @@
               </a>
             </td>
             <td class='text-center'>
+              {% if ops['id'] %}
```
r-vieira, line 89: consider replacing the square brackets in the `if` with .get('id'). [] will throw if the key doens't exist. .get() will return default value None (safer).

src/modules/site-v2/templates/tools/pairwise_indel_finder/submit.html
```diff
@@ -392,11 +392,14 @@ <h1 class="modal-title fs-5" id="generate-primers-modal-label">Thank You</h1>
...
-          console.error("Failed to retrieve divergent regions track.", err);
+          console.error(`Failed to retrieve divergent regions track: ${err.status} (${err.statusText})`);
         })
 
       // Update browser tracks
-      update_browser_tracks(browser, results);
+      update_browser_tracks(browser, []);
+    })
+    .catch((err) => {
+      console.log('Failed to update strains: ', err);
```
r-vieira, line 402: consider using console.error() for errors.

src/pkg/caendr/caendr/services/sql/etl/strain_annotated_variants.py
```diff
@@ -47,41 +51,70 @@ def parse_strain_variant_annotation_data(species, sva_fname: str):
...
       yield {
-        'id': idx,
-        'chrom': row[0],
-        'pos': int(row[1]) if row[1] else None,
-        'ref_seq': row[2] if row[2] else None,
-        'alt_seq': row[3] if row[3] else None,
-        'consequence': consequence,
+        'id':                 start_idx + idx,
+        'species_name':       species.name,
+        'chrom':              row['CHROM'],
```
r-vieira, line 70: potential for normalization of access across three variants (`row`: [], .get(), get_row()).

r-vieira (approved):
Looks good. See minor comments (not required for soft-launch).

src/modules/site-v2/templates/_scripts/utils.js
vlagrassa: Fixed in b8e13a8ee4f0b4ae16270af6cd73f31412cd31dc

src/modules/site-v2/templates/admin/etl/list.html
vlagrassa: Fixed in 75272b5ae23552b24186cdd111b2b27fafe7f6c8

src/modules/site-v2/templates/tools/pairwise_indel_finder/submit.html
vlagrassa: Fixed in dcced83fc3aca9224a6b1813011090c0dcb44de7

PR #222, opened 2023-05-24
ind strains filtering fix

orzut:


src/modules/site-v2/templates/strain/catalog.html
```diff
@@ -330,11 +330,14 @@ <h2 class="accordion-header" id="mappingSets">
...
-            $("#copyPasteStrains").val('')
+            $("#copyPasteStrains, #search").val('')
+
+
             if (lost_strains.length > 0) {
                 set_alert(`The following strain(s) were not found ${lost_strains.join(", ")}`)
             }
-
+            
+            filtered_by_species.show()
```
r-vieira: minor note -- the order of operations might be relevant here -- try swapping .show() and .sort() so that .show() renders the sorted content (this might avoid a quick UI flash of content being updated after shown).

r-vieira (changes requested):
looks good. See note about order of operations and adjust as needed.

src/modules/site-v2/templates/strain/catalog.html
```diff
@@ -347,6 +346,7 @@ <h2 class="accordion-header" id="mappingSets">
                 $('#indStrainsTable > tbody').append(row)
             })
             $('#checkoutBox').prop('disabled', !$("input:checkbox").is(":checked"));
+            filtered_by_species.show()
```
r-vieira, line 349: perfect!

r-vieira (approved):
looks good!

PR #225, opened 2023-05-24
Fix DB operations

vlagrassa:
- A few fixes for db_operations module
- Remove `Gene:` prefix from gene IDs, which should fix table rebuilding error


r-vieira (approved):
looks good!

PR #232, opened 2023-05-30
Fix/copy updates

natalieroman:
Changed the genome browser start text to changed based on device (desktop v. mobile). Removed duplicate paragraph on the heritability calculator. Removed unique pipeline users from stats page.


r-vieira (approved):
looks good

PR #239, opened 2023-06-01
Fix gene searching

vlagrassa:


src/modules/site-v2/templates/_scripts/gene-search.js
```diff
@@ -7,15 +7,22 @@
  */
 function prep_gene_search(gene, table_id, loading_id) {
 
+  const no_results_div = $(`${table_id}-no-results`)
```
r-vieira, line 10: technicality -- `table_id` here is being used as a selector instead of an id.

src/modules/site-v2/templates/_scripts/gene-search.js
```diff
@@ -55,17 +64,26 @@ function run_gene_search(gene, species, table_id, loading_id, callback) {
     contentType: 'application/xml',
   })
   .done(function(results) {
+    if (results.length > 0) {
+
+      // Map each gene to a set of cell values, and add to the table as <td> elements (if applicable)
+      $.each(results, (i, row) => {
```
r-vieira, line 70: whenever possible, avoid jQuery and prefer the built-in iterators. Eg: results.forEach((el, index) => {});

r-vieira (commented):
looks good

r-vieira (commented):
looks good

PR #240, opened 2023-06-01
Change file formats from TSV to CSV

vlagrassa:
Making a PR now to expedite merging, but some changes still coming.


src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
```diff
@@ -249,9 +250,17 @@ def submit():
 
 # TODO: Move internals of this to a service function
 @pairwise_indel_finder_bp.route("/report/<id>")
-@pairwise_indel_finder_bp.route("/report/<id>/tsv/<filename>")
+@pairwise_indel_finder_bp.route("/report/<id>/download/<file_ext>")
```
r-vieira, line 253: subjective API design -- putting on the RESTful hat here, the file_extension might be cleaner as a query param instead of part of the path since it is the same resource, just a different expression. Eg: /download?format=csv

r-vieira (approved):
looks good

PR #242, opened 2023-06-01
Update weekly visits graph styling

vlagrassa:
Makes background white, shows hover text value (like in strains graph).


r-vieira (approved):
great!

PR #246, opened 2023-06-05
Use new browser track & FASTA filepaths

vlagrassa:
Updated to match new values in rosetta stone document.

Currently, two copies of each file exist in GCP, one with the old name and one with the new name.  Once all code is using the new names, we can delete the old files.

Also adds fallback text if Release Notes file can't be loaded.


r-vieira (approved):
looks good

PR #247, opened 2023-06-05
Fix/strain colors

natalieroman:
Updated map pins and "Strains collected over time" chart with new colors


r-vieira (approved):
looks good

PR #251, opened 2023-06-06
Enhancement/homepage loading

orzut:


src/modules/site-v2/base/views/primary.py
```diff
@@ -28,6 +28,12 @@ def primary():
     'strain_listing': strain_listing,
   })
 
+@primary_bp.route('/strains-json')
```
r-vieira, line 31: quick point -- we haven't discussed this explicitly, but when designing a RESTful APIs, each 'url' maps to a 'resource'. The format (eg: .json) is a modifier to the request. 

Consider having the url be `/strains` default to serving in the `.json` format so that these return the exact same response: 
```json
/strains => [{...}]
/strains?format=json => [{...}]
```
Later, if we end up supporting other formats, folks will need to add a `?format=xml` to the URL.

r-vieira (approved):
looks good!

PR #252, opened 2023-06-06
Fix/mapping sets availability

natalieroman:
Removed unavailable sets and added availability message from Erik


r-vieira (approved):
approved

PR #257, opened 2023-06-07
Minor Tool Fixes

vlagrassa:
Biggest change is attempting to exclude PipelineOperation property `metadata` from indexes, to fix pipeline bug.  Will have to merge and re-deploy to ensure this works properly.

Should fix other minor issues, including:
- UTF-8 encoding bug in tool file uploads
- Indel Finder start/stop inputs (?)
- Indel Finder strain lists not always populating
- Minor text updates


src/modules/site-v2/base/utils/tools.py
```diff
@@ -71,7 +73,7 @@ def upload_file(request, filename, valid_file_extensions=None):
 
   # Match the file extension by splitting on right-most '.' character
   try:
-    file_ext = file.filename.rsplit('.', 2)[1]
+    file_ext = file.filename.rsplit('.', 1)[1]
```
r-vieira, line 76: consider using built-in:
```python
import pathlib
file_extension = pathlib.Path(file.filename).suffix 
```

r-vieira (approved):
looks good.

PR #262, opened 2023-06-07
Fix PipelineOperation exclude_from_indexes

vlagrassa:


src/pkg/caendr/caendr/models/datastore/pipeline_operation.py
```diff
@@ -4,7 +4,7 @@
 
 class PipelineOperation(Entity):
   kind = 'pipeline_operation'
-  exclude_from_indexes = ('metadata')
+  exclude_from_indexes = ('metadata',)
```
r-vieira, line 7: nice catch

r-vieira (approved):
approved

PR #271, opened 2023-06-09
Bugfix/tools heritability

r-vieira:


src/pkg/caendr/caendr/models/datastore/container.py
```diff
@@ -19,11 +19,12 @@ def get_props_set(cls):
       'repo',
       'container_name',
       'container_registry',
+      'container_repo',
       'container_tag',
     }
 
   def uri(self):
-    s = f"{self['repo']}/{self['container_name']}"
+    s = f"{self['container_repo']}/{self['container_name']}"
```
r-vieira, line 27: @r-vieira note: we want to skip the '/' if the container_repo is an empty string (in case it uses docker hub).

PR #276, opened 2023-06-09
Fix/tools fullscreen

natalieroman:
- Fixed "close fullscreen" button to be hidden when the escape key is used to exit fullscreen
- Hide the "view fullscreen" button for iPhones, as iOS doesn't support fullscreen api 
- Set background of tools to white so that they display correctly in dark mode.


r-vieira (approved):
looks great

PR #277, opened 2023-06-09
Changed alert class from "danger" to "warning"

natalieroman:
Changed the shipping notice color to "warning" to be more appropriate


r-vieira (approved):
approved

PR #285, opened 2023-06-12
Create Nemascan proxy module + script

vlagrassa:


src/modules/nemascan-proxy/publish_container.sh
```diff
@@ -0,0 +1,62 @@
...
+set +a
+source ../../../env/${ENV}/global.env
+
+if [ -n "${reload_git}" ]; then
+    rm -rf ./NemaScan
+fi
+git clone --depth 1 git@github.com:northwestern-mti/NemaScan.git
+
+
+docker build --no-cache -t gcr.io/${gcp_project_id}/nemascan-nxf:${tag} -f NemaScan/Dockerfile ./NemaScan
```
r-vieira: add flag `--platform linux/amd64` (otherwise the build breaks on some Apple M1s

src/modules/nemascan-proxy/publish_container.sh
```diff
@@ -0,0 +1,62 @@
...
+docker push gcr.io/${gcp_project_id}/nemascan-nxf:${tag}
```
r-vieira: make this line (push) conditional on the previous one (concatenate both with &&), or set a bash flag `-e` to exit if any errors are found.
```bash
set -e
```

src/modules/nemascan-proxy/publish_container.sh
```diff
@@ -0,0 +1,62 @@
...
+# Get current environment
+if [ -z ${ENV+x} ]; then
+    echo "Must set var ENV";
+    exit 1;
+else
+    echo "ENV is set to '$ENV'";
+fi
+
+# Map environment to GCP project
+if [ $ENV = "qa" ]; then
```
r-vieira: can we refactor this to use the PROJECT_ID values from /env/$ENV/global.env instead?

src/modules/nemascan-proxy/publish_container.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
```
r-vieira, line 2: add `set -e` (exits the script if any of the commands fails)

r-vieira (changes requested):
see inline comments

src/modules/nemascan-proxy/publish_container.sh
vlagrassa: Fixed in a73a67f3ba97a605e26034f645fba96b9453643a

src/modules/nemascan-proxy/publish_container.sh
vlagrassa: Added `-e` in a73a67f3ba97a605e26034f645fba96b9453643a

src/modules/nemascan-proxy/publish_container.sh
vlagrassa: Added in 8278f7ef4d8ff5dc0d537f119358a5cbd942855a

src/modules/nemascan-proxy/publish_container.sh
vlagrassa: Fixed in c0a931b4114f4f615548fd9fbb2f2231f31fa297

vlagrassa:
All comments should be fixed

src/modules/nemascan-proxy/publish_container.sh
```diff
@@ -0,0 +1,61 @@
...
+    exit 1;
+fi
+echo "Publishing with tag '${tag}'";
+
+
+# Clone git project
+if [ -n "${reload_git}" ]; then
+    rm -rf ./NemaScan
+fi
+git clone --depth 1 git@github.com:northwestern-mti/NemaScan.git
```
r-vieira, line 56: consider moving the GIT REPO URL to an ENV variable in global.env (add new section for NemaScan).

r-vieira (approved):
looks good. See note about removing the hardcoded GIT url and moving into global.env for each ENV. It is likely going to remain the same value across environments, but it will give us flexibility moving forward if we need to test a different source in one of the ENVs.

src/modules/nemascan-proxy/publish_container.sh
vlagrassa: Fixed in 8e844a0f2677272dbc7510e142423973128ba8b1 (PR #291)

vlagrassa:
Removed hardcoded Git & container information in #291.

PR #286, opened 2023-06-12
Fix/donation form

natalieroman:
Removed erroneous dollar sign and styled leading dollar sign. Added custom validation message in case user enters decimal currency values.


r-vieira (approved):
neat!

PR #287, opened 2023-06-12
Removed hamburger border

natalieroman:
Removed border so as to not appear as though icon is bumping up against edges of screen.


r-vieira (approved):
great

PR #289, opened 2023-06-12
Fixed margins on isotype page

natalieroman:
Fixed margins on the isotype page to be consistent with the rest of the site.


r-vieira (approved):
approved

PR #290, opened 2023-06-13
Fix/navbar breadcrumb margins

natalieroman:
Adjusted margins for navbar and breadcrumbs for mobile so that navbar takes up less space.


r-vieira (approved):
approved

PR #291, opened 2023-06-13
Read GitHub & container info from env files (nemascan-proxy) 

vlagrassa:


r-vieira (approved):
approved. Side note -- I keep thinking of the `make container` and `make publish` commands we have available for each module under `/src/modules/*`, and keep wondering if we should just go ahead and do the same for heritability-proxy and nemascan-proxy....what do you think?

PR #292, opened 2023-06-13
Fix/footer alignment

natalieroman:
Fixed "feedback" button to be pinned to top of footer on mobile/desktop. Fixed "Cite CeNDR" alignment in footer.


r-vieira (approved):
approved

PR #296, opened 2023-06-13
Removed placeholder text

natalieroman:
Removed placeholder text from the Order checkout form


r-vieira (approved):
approved. One day we will find out who's number and address they belong to

PR #300, opened 2023-06-14
Fix/mobile map

natalieroman:
Hid zoom controls on mobile. Made the species select toggle-able on mobile. Added the pin key to the species select. 

It might be my imagination, but I think the changes that I made to make the species select toggle-able are slowing down the pin loading on the map. @orzut and @r-vieira , would appreciate you taking a look at this.


r-vieira:
No worries, and thanks for the heads-up, @natalieroman. We'll look into it.

r-vieira (approved):
looks good

PR #304, opened 2023-06-14
Fix/heritability example data files

natalieroman:
Added example data files for different species to the "file information" modal


r-vieira:
Great!

PR #310, opened 2023-06-14
Fix/genetic mapping sample data

natalieroman:
Added sample data files for each species. Changed popover to modal.

### DIFF
--- a/src/modules/site-v2/base/views/tools/genetic_mapping.py
+++ b/src/modules/site-v2/base/views/tools/genetic_mapping.py
@@ -76,7 +76,9 @@ def genetic_mapping():
     ],
 
     # Sample data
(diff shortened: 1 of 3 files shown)


src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -76,7 +76,9 @@ def genetic_mapping():
     ],
 
     # Sample data
-    'sample_data_url': generate_blob_url(MODULE_SITE_BUCKET_ASSETS_NAME, 'data/nemascan_sample_data.tsv'),
+    'elegans_sample_data_url': generate_blob_url(MODULE_SITE_BUCKET_ASSETS_NAME, 'data/nemascan_sample_data_elegans.tsv'),
```
r-vieira, line 79: we will need to find a way to ensure that this file in GCP is always in-sync with the [newer] ones from Github.

r-vieira (approved):
approved

PR #315, opened 2023-06-14
updated fix_missing_thumbs.py file

orzut:


src/modules/img_thumb_gen/scripts/fix_missing_thumbs.py
```diff
@@ -23,7 +23,7 @@
 
 def generate_thumbnails(data, context):
   # Check if the file format is correct
-  if not data['name'].endswith('jpg') or not data['name'].endswith('jpeg'):
+  if not data['name'].endswith(('jpg', 'jpeg', 'JPG', 'JPEG')):
```
r-vieira: can we cover any case combinations? eg: convert the data['name'] to lower case and compare against it. Worry we might see Jpg or Jpeg, etc.

r-vieira (changes requested):
see note about case sensitivity

src/modules/img_thumb_gen/scripts/fix_missing_thumbs.py
orzut: pushed the changes in the commit [13c4906](https://github.com/AndersenLab/CAENDR/pull/315/commits/13c49060e1a4de8af314bba437123fa07da95836)

src/modules/img_thumb_gen/scripts/fix_missing_thumbs.py
```diff
@@ -23,7 +23,7 @@
 
 def generate_thumbnails(data, context):
   # Check if the file format is correct
-  if not data['name'].endswith(('jpg', 'jpeg', 'JPG', 'JPEG')):
+  if not data['name'].lower().endswith(('jpg', 'jpeg')):
```
r-vieira, line 26: great. Sry, should have mentioned to keep the '.' as part of the file extension (".jpg", ".jpeg").

r-vieira (approved):
approved

PR #325, opened 2023-06-15
Remove message container when alert close button is selected

natalieroman:
Fixed spacing issue when snackbar alerts are dismissed.


r-vieira (approved):
looks good

PR #326, opened 2023-06-15
Better error handling on isotype pages

vlagrassa:
- Doesn't abort the operation unless absolutely necessary (isotype can't be found)
- Handles strain list get/sort and image list get in separate try/except blocks
- If any filename fails to parse, ignores just that file

Still not sure what the source error is.  Sentry error message implies a problem with one of the `file.name.index()` function calls, but I didn't see any filenames in the production bucket that looked like they'd fail.  In the meantime, this fix should prevent the whole page from going down.

### DIFF
--- a/src/modules/site-v2/base/views/isotype.py
+++ b/src/modules/site-v2/base/views/isotype.py
@@ -71,9 +71,13 @@ def isotype_page(isotype_name, release=None):
     for s in isotype_strains:
       # Get images and thumbs for each strain
       for file in files:
-        start_idx = file.name.index('/')
-        end_idx = file.name.index('.')
-        file_name = file.name[start_idx+1:end_idx]
+        try:
+          start_idx = file.name.index('/')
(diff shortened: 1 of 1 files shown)


src/modules/site-v2/base/views/isotype.py
```diff
@@ -47,19 +47,37 @@ def isotype_page(isotype_name, release=None):
...
 
     image_urls = {}
     for s in isotype_strains:
       # Get images and thumbs for each strain
       for file in files:
-        start_idx = file.name.index('/')
-        end_idx = file.name.index('.')
-        file_name = file.name[start_idx+1:end_idx]
+        try:
+          start_idx = file.name.index('/')
```
r-vieira, line 75: @orzut -- instead of parsing this ourselves, can we use pathlib to extract the filename and extension?

r-vieira (changes requested):
@orzut please see inline note.

PR #328, opened 2023-06-15
Feature/new design request strains

orzut:


src/modules/site-v2/base/views/isotype.py
```diff
@@ -50,29 +51,27 @@ def isotype_page(isotype_name, release=None):
...
     for s in isotype_strains:
       # Get images and thumbs for each strain
       for file in files:
-        start_idx = file.name.index('/')
-        end_idx = file.name.index('.')
-        file_name = file.name[start_idx+1:end_idx]
-        if s.strain != file_name:
+        if s.strain not in file.name:
           continue
         else:
```
r-vieira, line 81: side note -- when we use early-return-guards, we can skip the `else:` and unindent the following code. 
eg: 
```python
list = []
for it in [...]:
  if it not in list:
    continue
  # skip else
  # do other stuff at this indentation level. 
```

src/modules/site-v2/base/views/isotype.py
```diff
@@ -50,29 +51,27 @@ def isotype_page(isotype_name, release=None):
...
 
     image_urls = {}
     for s in isotype_strains:
       # Get images and thumbs for each strain
       for file in files:
-        start_idx = file.name.index('/')
-        end_idx = file.name.index('.')
-        file_name = file.name[start_idx+1:end_idx]
-        if s.strain != file_name:
+        if s.strain not in file.name:
```
r-vieira: do we need to worry about case sensitivity here?

r-vieira (approved):
good

PR #332, opened 2023-06-15
Speed up data release page loading

vlagrassa:
Grab all filenames in one call, then check if each file exists in set.


r-vieira (approved):
💪

PR #334, opened 2023-06-16
Fix/submit strain link

natalieroman:
Changed "Submit A Strain" link in top nav and footer to open Google form in new tab.


r-vieira (approved):
looks good

PR #335, opened 2023-06-16
Templatize example data, improve Species entity fields

vlagrassa:
Note: Updates field names used in `Species` and `DatasetRelease` entities. These fields have all been copied to their new names in all environments, for zero-downtime transition.

Sample Files
- Move heritability & nemascan sample data file names to env var template
- Improved error reporting for invalid tokens in a `TokenizedString` (lists the token and whether it's in the replacement set or the template string)

Hyper-Divergent Regions & Browser Track Files
- Update "Hyper-Divergent Regions" section in dataset release to use new GCP location of track BED file
- Only show ^^ if defined for that species & release
- Move `browser_tracks` array from Species to DatasetRelease, so it can be updated for new tracks without altering older releases

Species Release Tracking
- Standardize names of Species release fields to use `release_` prefix


r-vieira (approved):
great!

PR #339, opened 2023-06-16
Allow user to re-run reports that ended in ERROR

vlagrassa:
Includes better checking for whether a job is:
- `finished`: `COMPLETE` or `ERROR`
- `not_err`: `SUBMITTED`, `RUNNING`, or `COMPLETE`

### DIFF
--- a/src/modules/site-v2/base/utils/tools.py
+++ b/src/modules/site-v2/base/utils/tools.py
@@ -135,7 +135,7 @@ def try_submit(EntityClass, user, data, no_cache):
     return {
       'cached':    True,
       'same_user': True,
-      'ready':     ex.report['status'] == TaskStatus.COMPLETE,
+      'ready':     TaskStatus.is_finished( ex.report['status'] ),
       'data_hash': ex.report.data_hash,
       'id':        ex.report.id,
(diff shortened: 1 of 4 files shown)


vlagrassa:
Specifically, if the data is resubmitted, the site will spin up a new `Task` for it.

src/modules/site-v2/base/utils/tools.py
```diff
@@ -135,7 +135,7 @@ def try_submit(EntityClass, user, data, no_cache):
     return {
       'cached':    True,
       'same_user': True,
-      'ready':     ex.report['status'] == TaskStatus.COMPLETE,
+      'ready':     Task.is_finished( ex.report['status'] ),
```
r-vieira, line 138: see if we can add this behavior to the `report` object itself, so usage could be: `report.is_finished()`

r-vieira, line 138: there is something uneasy about using exceptions for data flow -- in this case, `ex.report` appears to be used for logging (which is OK), but it is then used to set the `ready` property. Maybe this requires refactoring it so that the `try` block includes the data needed for the `catch/except` block during the closest stackframe

r-vieira (approved):
approving for now, and looking forward to refactor.

PR #340, opened 2023-06-20
Make DatabaseOperation a subclass of JobEntity & UserOwnedEntity

vlagrassa:
Class now uses updated logic for tracking unique ID, container version, and owner's user account.

Fixes bug with `database_operation` using container version from `module.env` instead of from GCP.

`database_operation` entity no longer explicitly tracks user's email, since it can be retrieved using the `username` field.  The corresponding `Task` object, however, does still take the email address, so this does not interfere with how it currently sends the notification email.


r-vieira (approved):
Great!

PR #341, opened 2023-06-20
Fix/indel primer table print

natalieroman:
Fixed indel primer report tables so that they shrink down to fit page when printed. Styled tables to be easier to read.


r-vieira (approved):
I always appreciate a good `crosshair` cursor hint.

PR #343, opened 2023-06-20
Fix/tool columns wrapping

natalieroman:
Removed the max-height from the input column for Genome Browser and Variant Annotation because it was causing elements to overlap on screens with a 3:2 aspect ratio (e.g., Pixelbook).


r-vieira (approved):
looks good

PR #344, opened 2023-06-20
Routing

vlagrassa:
- Redirect all hosts to the `MODULE_SITE_HOST` value for the environment
- Password protect QA
- Add redirect for Isotype List URL in older Nemascan reports


src/modules/site-v2/application.py
```diff
@@ -336,3 +347,70 @@ def handle_db_exceptions(err):
...
+
+def register_request_handlers(app):
+
+  # Redirect all URLs that point to this application to the host value in MODULE_SITE_HOST
+  # Individual API calls may fail if request body not copied over(?), but these should only be sent
+  # after initial page load, which will redirect to the correct host first
+  @app.before_request
+  def handle_redirects():
+    if request.host != MODULE_SITE_HOST:
+      return redirect(request.scheme + "://" + MODULE_SITE_HOST + request.full_path, code=307)
```
r-vieira, line 360: side note: 307 is fine for testing, but we should look into 301 to make it permanent and cache-able by browsers.

r-vieira (approved):
looks great!

PR #345, opened 2023-06-20
Add haplotype download buttons to tab, ensure download links download

vlagrassa:


src/modules/site-v2/templates/data/v2.html
```diff
@@ -292,13 +292,13 @@ <h2>Datasets</h2>
                                                     <i>et al.</i></a></td>
                                             <td>
                                                 {% if files.get('haplotype_png') %}
-                                                <a href="{{ files.get('haplotype_png') }}"> haplotype.png </a>
+                                                <a href="{{ files.get('haplotype_png') }}?response-content-disposition=attachment"> haplotype.png </a>
```
r-vieira, line 295: this works. In the effort of adding less code to maintain, consider the attribute `download` of the `a` tag: https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/download

r-vieira (approved):
approved

PR #352, opened 2023-06-21
Enhancement/homepage loading

orzut:


src/modules/site-v2/templates/primary/map.html
```diff
@@ -165,11 +163,18 @@
...
 const bounds = L.latLngBounds(southWest, northEast);
 
-const map = L.map('map', {minZoom:2, maxBounds: bounds, scrollWheelZoom: false, zoomControl: false, attributionControl: false,}).setView([20, -100], 2);
+const map = L.map('map', {renderer: L.canvas(), 
+                          minZoom:2, 
+                          maxZoom: 10, 
+                          maxBounds: bounds, 
+                          scrollWheelZoom: false, 
+                          zoomControl: false, 
+                          attributionControl: false})
```
r-vieira, line 176: quick note -- we might be required to display the attributionControl (should show that we are using leaflet) for the map. If `attributionControl` is set to true, does it show "leaflet" anywhere in the map?

src/modules/site-v2/templates/primary/map.html
```diff
@@ -261,43 +264,40 @@
...
-
   document.querySelector('#strains').addEventListener('click', e => {
     if (e.target.tagName !== 'INPUT') return;
-    
     const input = e.target
     const value = input.value;
-    showStrains(value)
+    const layer = layers_obj[value]
+    if (input.checked) {
+      overlays.addLayer(layer)
```
r-vieira, line 294: will this add duplicate layers if someone clicks/unclicks/clicks the same species?

r-vieira (commented):
looks good! see inline notes.

PR #354, opened 2023-06-21
updated the map in the About page

orzut:


src/modules/site-v2/templates/about/about.html
```diff
@@ -133,50 +158,129 @@ <h2>CaeNDR goals</h2>
...
+  const data = {{ strain_listing|tojson|safe }}
 
   data.forEach(function(d) {
     if (d.latitude) {
-      const innerHTML = `
+      const popup_content = `
         <div>
           <ul style="list-style-type:none; padding-left:0px">
-            ${d.isotype ? '<li><b><a href="../strains/isotype/' + d.isotype + '">Isotype: ' + d.isotype + '</a></b></li>': ''}
+            ${d.isotype ? '<li><b><a href="../isotype/' + d.isotype + '">Isotype: ' + d.isotype + '</a></b></li>': ''}
```
r-vieira, line 220: prefer the backtick string interpolation since it is what we have in other places in the code base.

r-vieira (approved):
approved

PR #358, opened 2023-06-22
API Pipeline: Better logging & checks

vlagrassa:
- Add entity or task ID to most log messages, to make them easier to filter/search
- Check if `PipelineOperation` exists when getting
- Check if `kind` is undefined before using

Not sure if this will fix all issues going forward, but should hopefully make the problem easier to find.  Good track record so far, but reproducing the error in question is difficult.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -262,6 +262,36 @@
             }
         },
         {
+            "name": "Run Site-v2 (local API)",
+            "type": "python",
+            "request": "launch",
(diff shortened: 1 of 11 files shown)


vlagrassa:
Encountered an error when running locally:
`Pipeline Error: EOF occurred in violation of protocol (_ssl.c:2393)`

Closing the PR until I understand this better.

r-vieira (approved):
looks great!

PR #361, opened 2023-06-23
Db ops drop tables fix + preflight check

vlagrassa:
## Drop tables in reverse order

Deals with `ForeignKey` dependencies.

Currently, only relevant for `Gene` and `GeneSummary`, where the former has a `ForeignKey` column into the latter.  When handling tables one at a time, they have to be populated and cleared in reverse order.

## Preflight Check

Prevents submitting ETL db-ops jobs if any required files missing in GCP.


r-vieira (approved):
approved!

PR #364, opened 2023-06-30
Fix/map zoom

natalieroman:
Changed color of zoom buttons, hide zoom buttons on mobile, disabled zoom out button when at max zoom out.


r-vieira (approved):
looks great!

PR #367, opened 2023-07-07
Fix haplotype rendering

vlagrassa:
Differentiate "view in new tab" and "download" functionality with `force_download` util function.

### DIFF
--- a/src/modules/site-v2/templates/_scripts/utils.js
+++ b/src/modules/site-v2/templates/_scripts/utils.js
@@ -48,6 +48,9 @@ function form_data_to_object(form_id) {
 
 function force_download(el, url) {
 
+  // If this function has run on the element already, use the href normally
+  if (el.href[el.href.length - 1] != '#') return;
+
   // Fetch the provided URL and create a blob object from it
   fetch(url)
     .then(response => response.blob())
@@ -56,14 +59,11 @@ function force_download(el, url) {
       // Create an object URL for the blob object
       const url = URL.createObjectURL(blob);
 
-      // Create a new anchor element
-      const a = document.createElement('a');
+      // Point the element's href to the URL of the new blob object
+      el.href = url;
 
-      // Set the href and download attributes for the anchor element
-      a.href = url;
-      a.download = el.download || 'download';
-
-      a.click();
+      // Re-click the element, now that the href is updated
+      el.click();
     })
 
   // Stop the first click from following the dummy link
--- a/src/modules/site-v2/templates/_scripts/utils.js
+++ b/src/modules/site-v2/templates/_scripts/utils.js
@@ -44,3 +44,28 @@ function form_data_to_object(form_id) {
   });
   return data;
 }
+
+
+function force_download(el, url) {
+
+  // Fetch the provided URL and create a blob object from it
+  fetch(url)
+    .then(response => response.blob())
+    .then(blob => {
+
+      // Create an object URL for the blob object
+      const url = URL.createObjectURL(blob);
+
+      // Create a new anchor element
+      const a = document.createElement('a');
+
+      // Set the href and download attributes for the anchor element
+      a.href = url;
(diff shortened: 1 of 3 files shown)


src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -44,3 +44,28 @@ function form_data_to_object(form_id) {
   });
   return data;
 }
+
+
+function force_download(el, url) {
+
+  // If this function has run on the element already, use the href normally
+  if (el.href[el.href.length - 1] != '#') return;
```
r-vieira: prefer triple equals (=== and !==)

r-vieira: consider adding a console.warning() if conditions are not met.

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -44,3 +44,28 @@ function form_data_to_object(form_id) {
   });
   return data;
 }
+
+
+function force_download(el, url) {
```
r-vieira: Some thoughts on the utility interface: 

Looking at how the force_download() is being used, perhaps it can be simplified -- if the force_download() function is being used as an event handler callback, you can keep the `href` populated in the `<a/>` element, and just use `onclick=force_download`. This will call your function with a parameter `event` that contains a reference to the `<a/>` tag in `event.target` where you can find URL in `event.target.href`. At that point, since you have access to the `event` itself in the callback, you can use event.stopPropagation(), event.preventDefault() to avoid having it the browser follow the link. 

on the other hand, if we want a `pure` download function, we should pass the `url` as the only parameter, and then create/destroy the necessary <a/> tag dynamically (add to the body, click, remove from the body).

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -44,3 +44,28 @@ function form_data_to_object(form_id) {
...
+
+  // If this function has run on the element already, use the href normally
+  if (el.href[el.href.length - 1] != '#') return;
+
+  // Fetch the provided URL and create a blob object from it
+  fetch(url)
```
r-vieira: a caveat of having this be part of an event handler, is that if there are any issues fetching the blob (eg: cors, missing auth, etc), there is no UI warning for the end-user. How can we expose error-states to the front-end to avoid the end-user running into the situation "clicked on a button/link and nothing happened"? Maybe adding a banner notification about the failed attempt to download the file, or, since we have access to the </a> element, if there is a known "component" around the download link that includes a <span class='error'/> you might be able to navigate around the </a> to find the error tag and make it visible, etc.

r-vieira (changes requested):
this works, but added a few thoughts on simplifying the interface.

src/modules/site-v2/templates/_scripts/utils.js
vlagrassa: Currently, I don't think we have any way to signal a failed download to the user... might be worth looking into for more than just this link.

(If the file doesn't exist in GCP, for example, the user gets a `NoSuchKey` XML error page.)

vlagrassa:
Thank you for the suggestions!  I've revised the interface:
```
<a href="..." onclick="force_download(event)" download="filename.png"> ... </a>
```

I wasn't able to get rid of the `event` argument to `force_download`, but I think this is much better.

I also made the function safer:
- Creates a temporary `a` element for the initial click, so the function doesn't recursively trigger itself.  If anything goes wrong, it won't be able to create an endless loop.
- Checks for empty `href` attribute and throws an error instead of proceeding.

I'm not sure yet how to handle errors... I don't think any of the links in the download page have good error handling.  Might be worth looking to in its own branch.  I'll add a card to Basecamp.

r-vieira (approved):
approving the change to go live for now.

PR #368, opened 2023-07-07
Fix BAM/BAI downloads

vlagrassa:
Fixes bulk download script and individual download links.

- BAM/BAI download script is now generated on-demand under `data_downloads_bp.download_bam_bai_script`
- Download script now depends on species and release
- `BAM_BAI_DOWNLOAD_SCRIPT_NAME` and `BAM_BAI_PREFIX` are now templated environment variables, and rely on species parameter(s)
- Individual downloads now take a species parameter (makes new species available)

The original maintenance function that would generate the script daily has been updated to use new generator function, though it might be obsolete at this point.  I suspect it was originally done this way because generating the full list of signed URLs took a long time, so this would make a copy of the file available for immediate download.  Since we're no longer using signed URLs, this is no longer a concern.

Is there any use case where we would want to generate the download script (using either signed or unsigned URLs) and upload to GCP?


r-vieira (approved):
looks good

PR #371, opened 2023-07-12
Small Dataset Release Fixes

vlagrassa:
- Include `transposon_calls.bed` in `caendr` GCP project(s), and show download link / not included message like other files
- Make "not included" message styling match for button links and non-button links
- Fix: Hide `Haplotypes` tab unless PNG exists (was checking for PDF, based on older layout)
- Make macro to render download link / not included message
- Add standard prefix to haplotype & swept haplotype files


r-vieira (approved):
approved

PR #372, opened 2023-07-13
Small Gene Search Fixes

vlagrassa:
- Fix back-end sorting of results:
  - Use a more natural sorting algorithm (e.g. `lin-2` before `lin-10`)
  - Sort before truncating (taking top 10), so most relevant results actually appear
- If query is just species-specific gene prefix, search for it directly instead of removing it -- otherwise query would match all genes (equivalent to searching for `""`).  Otherwise, the prefix is removed from the search.
- Display error message in dropdown if query fails (distinct from empty results)
- Add small margins around dropdown messages, so they look less awkward

### DIFF
--- a/src/modules/site-v2/templates/_scripts/igv-browser.js
+++ b/src/modules/site-v2/templates/_scripts/igv-browser.js
@@ -69,7 +69,9 @@ function add_track(track_name, track_data = null) {
   if (!tracks.has(track_name)) {
     const track = track_data || get_track(track_name)
     tracks.add(track_name);
-    return igv.getBrowser().loadTrack(track);
+    return igv.getBrowser().loadTrack(track).catch(error => {
+      console.log('Error loading track ' + track_name + ': ', error)
+    });
   }
 }
 
--- a/src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
+++ b/src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
@@ -69,10 +69,10 @@
                 </thead>
                 <tbody id="orthologs"></tbody>
               </table>
-              <div id="g-search-table-empty" style="display:none;">
+              <div id="g-search-table-empty" class="m-2" style="display:none;">
                 That gene name is not found in this species, so either that gene does not exist in this species or a different gene name is indexed.
               </div>
-              <div id="g-search-table-error" style="display:none;">
+              <div id="g-search-table-error" class="m-2" style="display:none;">
                 There was an error processing your gene search.
               </div>
             </div>
--- a/src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
+++ b/src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
@@ -69,7 +69,7 @@
                 </thead>
                 <tbody id="orthologs"></tbody>
               </table>
-              <div id="g-search-table-no-results" style="display:none;">
+              <div id="g-search-table-empty" style="display:none;">
                 That gene name is not found in this species, so either that gene does not exist in this species or a different gene name is indexed.
               </div>
             </div>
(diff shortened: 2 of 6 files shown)


src/modules/site-v2/templates/_scripts/igv-browser.js
```diff
@@ -69,7 +69,9 @@ function add_track(track_name, track_data = null) {
   if (!tracks.has(track_name)) {
     const track = track_data || get_track(track_name)
     tracks.add(track_name);
-    return igv.getBrowser().loadTrack(track);
+    return igv.getBrowser().loadTrack(track).catch(error => {
+      console.log('Error loading track ' + track_name + ': ', error)
```
r-vieira, line 73: should this error be exposed to the end-user?

src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
```diff
@@ -236,6 +239,14 @@ <h1 class="modal-title fs-5" id="infoModalLabel">Track Information</h1>
   {% for track in default_tracks %} {% if track.checked %} "{{ track.name }}", {% endif %} {% endfor %}
 ];
 
+// Divs used by the gene search
+const gene_search_div_ids = {
```
r-vieira: minor semantic/naming detail -- consider using `..._element_...` instead of `..._div_...`. E.g. `gene_search_element_ids = `

src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
```diff
@@ -236,6 +239,14 @@ <h1 class="modal-title fs-5" id="infoModalLabel">Track Information</h1>
   {% for track in default_tracks %} {% if track.checked %} "{{ track.name }}", {% endif %} {% endfor %}
 ];
 
+// Divs used by the gene search
+const gene_search_div_ids = {
+  "table":   "#g-search-table",
```
r-vieira, line 244: semantic detail -- if these are `ids`, then they shouldn't include the `#` in the values. On the other hand, if these are `selectors`, then the `#` should be included.

r-vieira (changes requested):
see inline comments (main one is about exposing the error to the end-user)

src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
vlagrassa: Fixed in https://github.com/AndersenLab/CAENDR/pull/372/commits/7031c98c648bc623a016dd431881c4189573db16

src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
vlagrassa: Fixed in https://github.com/AndersenLab/CAENDR/pull/372/commits/7031c98c648bc623a016dd431881c4189573db16

src/modules/site-v2/templates/_scripts/igv-browser.js
vlagrassa: Good question... if there's an error getting the track, the IGV Browser will show a pop-up with the full URL.

Might be helpful for us to log somewhere if there's an error?  Or just ignore it?

r-vieira: Got it. In this case, we want to capture it in Sentry.IO. (By default SentryIO will track JS exceptions at the top level. SentryIO might also have helper functions that let you track a single event, if that helps). In short, we should be able to find this error in SentryIO and use automated notifications so we are aware of it.

r-vieira:
See note about having SentryIO capture the error (missing IGV track) so we can automate handling on our side.

PR #373, opened 2023-07-14
Fix vbrowser tool output / download file, strain list download file

vlagrassa:


src/modules/api/pipeline-task/test.py
```diff
@@ -4,7 +4,7 @@
 from caendr.models.task import HeritabilityTask
 
 from caendr.services.heritability_report import HERITABILITY_TASK_QUEUE_NAME, create_new_heritability_report
-from caendr.utils.data import convert_data_table_to_tsv, get_object_hash, unique_id
+from caendr.utils.data import convert_data_table_to_csv, get_object_hash, unique_id
```
r-vieira: Consider exposing a more generic interface to the data conversion, eg: `convert_data_table(..., format='csv')`

src/modules/site-v2/base/views/strains.py
```diff
@@ -95,13 +95,19 @@ def external_links():
 #
 # Strain Data
 #
-@strains_bp.route('/download/<release_name>/<species_name>/strain-data/<file_ext>')
+@strains_bp.route('/download/<species_name>/<release_name>/strain-data/<file_ext>')
```
r-vieira, line 98: ensure that no other part of the site (or release bash download scripts) link to the old URL structure.

src/modules/site-v2/base/views/tools/variant_annotation.py
```diff
@@ -74,37 +76,57 @@ def variant_annotation():
...
+def query_position(species_name=None):
 
   # Extract the query
   payload = json.loads(request.data)
   query = payload.get('query')
 
+  # Get the species from the URL, allowing undefined
+  if species_name:
+    try:
+      species = Species.get(species_name.replace('-', '_'))
```
r-vieira: see if there is a better interface for getting the species from the URL that can be reused across all endpoints, eg: `Species.get_from_url(...)`.

src/pkg/caendr/caendr/models/sql/strain.py
```diff
@@ -52,3 +52,64 @@ def to_sortable_strain(strain):
...
+    col_list = [ c for c in col_list if c[0] is not None]
+
+    # Sort by index in ordered list, and map back to just the column object
+    col_list = [ c[1] for c in sorted(col_list, key=lambda x: x[0]) ]
+
+    # Return names or full columns, as requested
+    # Can't just return ordered list above because some of the columns listed are not defined in this table
+    if names_only:
+      return [ c.name for c in col_list ]
+    else:
```
r-vieira: you can remove `else` and de-indent the last return

src/pkg/caendr/caendr/models/sql/strain_annotated_variant.py
```diff
@@ -95,20 +93,21 @@ def column_default_visibility(col):
...
+  def generate_interval_sql(cls, interval, species=None):
     interval = interval.replace(',','')
     chrom = interval.split(':')[0]
     range = interval.split(':')[1]
     start = int(range.split('-')[0])
     stop = int(range.split('-')[1])
 
-    query = f"SELECT * FROM {cls.__tablename__} WHERE chrom='{chrom}' AND pos > {start} AND pos < {stop};"
+    species_clause = f"species_name='{species.name}' AND " if species else ''
+    query = f"SELECT * FROM {cls.__tablename__} WHERE {species_clause} chrom='{chrom}' AND pos > {start} AND pos < {stop};"
```
r-vieira: this looks unsafe....can we build the query binding values instead of string concat?

r-vieira (changes requested):
Let's take this opportunity and review all the sql query formatted f-strings and convert to ORM query bindings.

src/modules/api/pipeline-task/test.py
vlagrassa: Fixed in https://github.com/AndersenLab/CAENDR/pull/373/commits/299f9c52187ba3822c883acf4bbc66a14b1c0f49

src/modules/site-v2/base/views/tools/variant_annotation.py
vlagrassa: Fixed in https://github.com/AndersenLab/CAENDR/pull/373/commits/2a427db1832c241399ea008eaea9c3fb52278d9b, with minor tweaks in following two commits

src/pkg/caendr/caendr/models/sql/strain.py
vlagrassa: Fixed in https://github.com/AndersenLab/CAENDR/pull/373/commits/2e4f1b4c1e274af8f341c0a78ff9011025073ea7

src/pkg/caendr/caendr/models/sql/strain_annotated_variant.py
vlagrassa: Fixed in https://github.com/AndersenLab/CAENDR/pull/373/commits/f69b630d07e43ff9898a96580f29df2a7ce5c2bd, with small updates in following 3 commits

vlagrassa:
All fixes should be applied -- let me know how this looks. Thanks!

src/modules/site-v2/templates/releases/download_tab_isotype_v1.html
```diff
@@ -54,7 +54,7 @@ <h4>VCF</h4>
                 Strain Data including isolation location, isotype information, and more is available for download.
                 <br />
                 <br />
-                <a href="{{ url_for('request_strains.strains_data_csv', release_name = release_version, species_name = species.name, file_ext='csv') }}" type="button" class="btn btn-default btn-block btn-success">Download Strain Data</a>
+                <a href="{{ url_for('request_strains.strains_data_csv', release_name = release_version, species_name = species.get_url_name(), file_ext='csv') }}" type="button" class="btn btn-default btn-block btn-success">Download Strain Data</a>
```
r-vieira: Could we find a more descriptive or intuitive name for `#get_url_name()` (eg: referring to prefix/suffix/slug/etc)?

src/pkg/caendr/caendr/models/sql/strain.py
```diff
@@ -41,14 +44,71 @@ def __repr__(self):
...
+      'isolated_by',
+      'sampling_date',
+      'sampling_date_comments',
+      'notes',
+      'strain_set',              # "set" in source data sheet
+      'issues',
+      'issue_notes',
+      'isotype_ref_strain',
+      'sequenced',               # "wgs_seq" in source data sheet
+      # 'peel-zeel',
```
r-vieira: remove unused/commented-out dependencies

src/pkg/caendr/caendr/models/sql/strain.py
```diff
@@ -41,14 +44,71 @@ def __repr__(self):
...
+
+    # Filter out any columns not in the ordered list
+    col_list = [ c for c in col_list if c[0] is not None]
+
+    # Sort by index in ordered list, and map back to just the column object
+    col_list = [ c[1] for c in sorted(col_list, key=lambda x: x[0]) ]
+
+    # Return names or full columns, as requested
+    # Can't just return ordered list above because some of the columns listed are not defined in this table
+    if names_only:
```
r-vieira: prefer returning consistent data shapes (col vs col.name) -- eg: in this method, only return cols. Add a new method that returns the col names (could call this one first, etc).

src/pkg/caendr/caendr/utils/bio.py
```diff
@@ -16,3 +19,62 @@ def arm_or_center(chrom, pos):
...
+
+    Returns:
+      dict:
+        - chrom (str): The chromosome
+        - start (int): The start location
+        - stop (int):  The stop location
+  '''
+
+  # Perform RegEx search on interval string
+  match = re.search(CHROM_INTERVAL_REGEX, interval.replace(',',''))
```
r-vieira, line 43: good candidate for unit test `CHROM_INTERVAL_REGEX`.

src/pkg/caendr/caendr/utils/bio.py
```diff
@@ -16,3 +19,62 @@ def arm_or_center(chrom, pos):
...
+
+  # If there was a match, parse the capture groups
+  if match:
+    return {
+      'chrom': match.group(1),
+      'start': int(match.group(2)),
+      'stop':  int(match.group(3)),
+    }
+
+  if not silent:
```
r-vieira: hum...we should review the behavior for `silent==True`, should it log something?

r-vieira (changes requested):
Looks great! See inline comments (minor details).

src/modules/site-v2/templates/releases/download_tab_isotype_v1.html
vlagrassa: Updated in [6617454](https://github.com/AndersenLab/CAENDR/pull/373/commits/6617454bc096bc8048042cab2a966f997abc6fe6) to `get_slug()`

src/pkg/caendr/caendr/models/sql/strain.py
vlagrassa: Removed in https://github.com/AndersenLab/CAENDR/pull/373/commits/738da22dab55e26c8f41da9b1dfc585970b418e9

src/pkg/caendr/caendr/models/sql/strain.py
vlagrassa: Fixed in https://github.com/AndersenLab/CAENDR/pull/373/commits/714c44ede5b1593590bded8331dd39f1741d4a76

src/pkg/caendr/caendr/utils/bio.py
vlagrassa: I'm inclined to say that if we want a log message, the caller should probably be using `silent=False` and logging the error itself.  That said, I can see why we might want to add a log message anyways, in case the caller doesn't do that.  Happy to add one if you think it would be useful.

--

A bit more broadly -- my idea for the parameter was to have a nice way to toggle between "just return `None`" and "raise an error" in case the pattern doesn't match.  Then, it would be up to the caller to determine which behavior is more useful.  Implicit in that would be that `silent=True` would indicate that a failed match isn't a reason to halt execution.

E.g. at the top level of the `variant_annotation.query_interval` view function, we return an empty response if the result is `None`, whereas in `run_interval_query`, the whole query should fail with the appropriate error if the interval can't be parsed.  That said, I can also see an argument for raising an error in both cases, and for catching it in the former case & letting it propagate in the latter.

r-vieira: Got it. If one squints, it might feel like we are trying to implement a workaround for throw/catch. Would it be reasonable to expect the function to throw (in some cases) and let the caller handle it with try/catch? At that point the caller can decide how it wants to proceed in case of an error. 

I'm not opposed to the extra `silent` argument in these functions; I am mainly looking for ways to simplify and reduce cognitive load throughout the codebase & function signatures.

lmk what you think.

vlagrassa: Looking back at how we use these functions, I think throw/catch like you suggested does make more sense.  Updated in https://github.com/AndersenLab/CAENDR/pull/373/commits/551b3abf14777bcf657e4a39be576e4b487d5101.

PR #374, opened 2023-07-14
Bump IGV Browser to 2.8.3

vlagrassa:


vlagrassa:
Setting as draft until we can discuss how far we should bump the library.

src/modules/site-v2/templates/tools/pairwise_indel_finder/submit.html
```diff
@@ -135,7 +135,7 @@ <h1 class="modal-title fs-5" id="generate-primers-modal-label">Thank You</h1>
 let strains;
 
 /* IGV Browser */
-const browser_div = $("#browser");
+const browser_div = $("#browser")[0];
```
r-vieira: semantics -- consider calling this `browser_element` (or `browser_domElement`) instead of implementation specific detail `_div`. 
Also, whenever possible, we should minimize dependencies on jQuery (`$`). In this case, (as long as js is running after the DOM is loaded), we can use the HTML5 equivalent, `document.body.querySelector(...)` (preferred).

src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
```diff
@@ -432,7 +432,7 @@ <h1 class="modal-title fs-5" id="infoModalLabel">Track Information</h1>
 
 function set_position(coord) {
   // Sets browser position based on input
-  igv.getBrowser().search(coord);
+  get_browser().search(coord);
```
r-vieira, line 446: add a guard here in case get_browser() is still uninitialized.

r-vieira (approved):

r-vieira (approved):
looks great.

PR #375, opened 2023-07-17
Fix IGV Browser Fullscreen

vlagrassa:
Any time fullscreen state changes, call `visibilityChange()` on the global IGV object (if it exists).  Should fix browser fullscreen behavior.


r-vieira (approved):
looks good.

PR #376, opened 2023-07-18
fixed search filter

orzut:
The source of the bug was incorrect id of the search input. Now it should worked as expected.


r-vieira (approved):
looks good.

PR #377, opened 2023-07-19
Minor Corrections

vlagrassa:
- Genome Browser Divergent Regions Summary info
- Enable all species in Request Strains -> strain sets dropdown (now that data exists on back end)
- Close both strain sets info dropdowns by default in Request Strains page


r-vieira (approved):
approved

PR #378, opened 2023-07-20
Bugfix: Submit correct VCF versions to tools

vlagrassa:
When starting a Nemascan or Heritability task, get the most recent release for the given species, and include that in the environment as `VCF_VERSION`.  Should fix bug when submitting job for new species.

Also includes a few extra checks when trying to start a `Task` (most notably, ensuring the `Task` corresponds w an existing `Entity`), as well as some re-organization.  I want to do some heavier re-organizing, but I'll wait until we're focusing on the containers.

### DIFF
--- a/src/modules/api/pipeline-task/pipelines/db_op.py
+++ b/src/modules/api/pipeline-task/pipelines/db_op.py
@@ -1,30 +1,27 @@
-import os
-import logging
-
 from caendr.utils import monitor
 from caendr.models.task import DatabaseOperationTask
 from caendr.models.datastore import DatabaseOperation
 from caendr.services.cloud.lifesciences import start_pipeline
 from caendr.models.lifesciences import ServiceAccount, VirtualMachine, Resources, Action, Pipeline, Request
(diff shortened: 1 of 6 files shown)


src/modules/api/pipeline-task/pipelines/db_op.py
```diff
@@ -1,30 +1,27 @@
-import os
-import logging
-
 from caendr.utils import monitor
 from caendr.models.task import DatabaseOperationTask
 from caendr.models.datastore import DatabaseOperation
 from caendr.services.cloud.lifesciences import start_pipeline
 from caendr.models.lifesciences import ServiceAccount, VirtualMachine, Resources, Action, Pipeline, Request
-from caendr.utils.json import get_json_from_class
```
r-vieira, line 9: this is being removed, but I'll add this note here anyway for future reference -- 
API semantic note -- `get_json_from_class` is usually implemented as a method `#to_json()` on the class that it applies to.

r-vieira (approved):
looks good

PR #381, opened 2023-07-25
Updated strain issues template to display "no issues found" message i…

natalieroman:
Updated the "strain issues" page template to display a "no issues found" message if there are no issues in the release. Previous template was rendering an empty table when there were no issues, which is bad for accessibility / users with screen readers.

### DIFF
--- a/src/modules/site-v2/templates/releases/download_tab_strain_v2_issues.html
+++ b/src/modules/site-v2/templates/releases/download_tab_strain_v2_issues.html
@@ -1,4 +1,12 @@
 
+{% set ns = namespace(hasIssue=false) %}
+{% for strain in strain_listing_issues %}
+    {% if strain.issues == True %}
+        {% set ns.hasIssue=true %}
+    {% endif %}
+{% endfor %}
+
+{% if ns.hasIssue == true %}
 <table class="table table-hover table-condensed strain-table">
     <thead>
         <tr>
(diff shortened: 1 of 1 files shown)


src/modules/site-v2/templates/releases/download_tab_strain_v2_issues.html
```diff
@@ -1,4 +1,12 @@
 
+{% set ns = namespace(hasIssue=false) %}
+{% for strain in strain_listing_issues %}
+    {% if strain.issues == True %}
+        {% set ns.hasIssue=true %}
+    {% endif %}
+{% endfor %}
+
+{% if ns.hasIssue == true %}
```
r-vieira: prefer the capitalized usage for boolean types (`True` and `False`). In this case, we can short-circuit as well, by only using:
```
{% if ns.hasIssue %}
```

src/modules/site-v2/templates/releases/download_tab_strain_v2_issues.html
```diff
@@ -1,4 +1,12 @@
 
+{% set ns = namespace(hasIssue=false) %}
```
r-vieira: prefer capitalized booleans (eg: `False`)

r-vieira (changes requested):
use capitalized booleans (`True`/`False`).

src/modules/site-v2/templates/releases/download_tab_strain_v2_issues.html
natalieroman: Just pushed changes to capitalize.

natalieroman:
@r-vieira Just pushed requested changes.

r-vieira (approved):
looks good

PR #383, opened 2023-08-01
Set up Pub/Sub messages with CloudRun Jobs

vlagrassa:
- Manually push a Pub/Sub message when a new Job is run.
- Minor bug fixes & reorganization
- Bump module version to `v1.1.0` -- I think a new minor revision is justified, since we've swapped out the back-end service we're using to run jobs

### DIFF
--- a/src/modules/api/pipeline-task/pipelines/utils.py
+++ b/src/modules/api/pipeline-task/pipelines/utils.py
@@ -86,7 +86,7 @@ def start_job(payload, task_route, run_if_exists=False):
   # Run the CloudRun job
   # TODO: Wait to make sure job is created first?
   try:
-    run_response = handler.run_job()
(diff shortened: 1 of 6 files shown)


src/modules/api/pipeline-task/pipelines/utils.py
```diff
@@ -95,7 +95,7 @@ def start_job(payload, task_route, run_if_exists=False):
   except HttpError as ex:
     if ex.status_code == 400:
       time.sleep(5)
-      run_response = handler.run_job()
+      run_response, pub_sub_id = handler.run_job()
```
r-vieira, line 98: thanks for the confirmation. An alternative is to use a `retry` library/decorator where you can specify the # of retries, wait time, backoff scale, etc.

r-vieira (approved):
looks good!

PR #384, opened 2023-08-03
Improve error handling on About & Data Release pages + minor text updates

vlagrassa:
Original title: _Fix "Global distribution" section text on About page_

Makes the About page more robust if strain list can't be retrieved from database.  In the future, we may want to handle that error on the page explicitly.

Adds fallback behavior to the Data Release page if GCP files are unavailable.  Also removes dependence on Strain table, which wasn't being used.

Small text updates for About page "Global distribution" section.

### DIFF
--- a/src/modules/site-v2/base/views/data/releases.py
+++ b/src/modules/site-v2/base/views/data/releases.py
@@ -10,6 +10,7 @@ from flask import (request,
                     Blueprint,
                     send_file,
                     url_for,
+                    flash,
                     abort)
 from caendr.services.logger import logger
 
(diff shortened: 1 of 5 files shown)


src/modules/site-v2/base/views/data/releases.py
```diff
@@ -98,7 +99,12 @@ def data_release_list(species, release_version=None):
   }
 
   # Get list of files based on species
-  files = release.get_report_data_urls_map(species.name)
+  try:
+    files = release.get_report_data_urls_map(species.name)
```
r-vieira, line 103: minor semantics: consider renaming `files` to `files_map` or `files_dict`.

r-vieira (approved):
approved

PR #385, opened 2023-08-07
Fix: Email notification being sent too early

vlagrassa:
Fixes bug in checking for CloudRun job execution. Original check for `done` was too lax, meaning the email would be sent well before the job was done.

### DIFF
--- a/src/pkg/caendr/caendr/services/cloud/cloudrun.py
+++ b/src/pkg/caendr/caendr/services/cloud/cloudrun.py
@@ -100,7 +100,7 @@ def get_job_execution_status(name):
 
   # TODO: Make sure this is adequate
   for condition in response['conditions']:
-    if condition['type'] == 'Completed':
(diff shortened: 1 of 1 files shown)


src/pkg/caendr/caendr/services/cloud/cloudrun.py
```diff
@@ -100,7 +100,7 @@ def get_job_execution_status(name):
 
   # TODO: Make sure this is adequate
   for condition in response['conditions']:
-    if condition['type'] == 'Completed':
+    if condition['type'] == 'Completed' and condition['state'] == 'CONDITION_SUCCEEDED':
```
r-vieira, line 103: whenever possible, prefer class ENUMS instead of literal strings for logic control. Eg: `Status.COMPLETED` instead of `'Completed'`. (applies to `'CONDITION_SUCCEEDED'` as well.

r-vieira (approved):

PR #386, opened 2023-08-08
Address scheduling/connection errors (CloudRun create/run scheduling, SSLEOFError broken pipe error)

vlagrassa:
- Add the `backoff` package v2.2.1 ([link](https://pypi.org/project/backoff/)):
  - Add delay & retries between CloudRun API `create` call and `run` call, to ensure the job is started correctly
  - Retry if initial `start_job` request encounters an `SSLEOFError`, which tends to happen on the first request after a long period of requests.  Manually starting a dummy job, letting it fail, and starting a new job has usually worked so far, so this formalizes that retry process.
- Explicitly close all db connections on app context teardown. Will hopefully prune connections before they go stale, which could ideally avoid the above problems altogether.  (That said, I'm not sure the above is a db connection issue...)


r-vieira (approved):
looks good

vlagrassa:
Quick update: the catch-and-retry on `SSLEOFError` successfully caught and handled the broken pipe error in a new ECHO job in the QA environment.  This doesn't address the root problem, but it should stop jobs from auto-failing going forward.

PR #387, opened 2023-08-09
Fix Pub/Sub status checks (and maybe SSLEOFError once and for all)

vlagrassa:
Explicitly ack or nack a Pub/Sub message sent to the `status` endpoint, using codes `200` and `100` respectively.  This means the subscription will continue to send periodic updates until the job finishes.

Involves changing the configuration of the Subscription to wait between messages, using exponential backoff.  I tested my proposed change by manually updating Pub/Sub in GCP.  Encoding this change in Terraform would formalize it.

Also, adds a `use_service` decorator which creates a Google Client API resource on-the-fly before the function is called, and closes all connections once it finishes.  My working theory is that the `SSLEOFError` comes from stale resource connections, and that closing them will solve this issue.  Initial tests look promising, but I'll test this again tomorrow to really get a cold start scenario.

### DIFF
--- a/src/modules/api/pipeline-task/routes/task.py
+++ b/src/modules/api/pipeline-task/routes/task.py
@@ -11,7 +11,7 @@ from caendr.models.task import TaskStatus
 from caendr.models.pub_sub import PubSubAttributes, PubSubMessage, PubSubStatus
 
 from caendr.services.cloud.task import update_task_status, verify_task_headers
-from caendr.services.cloud.pubsub import get_attribute, make_pubsub_response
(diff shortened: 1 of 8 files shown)


vlagrassa:
I have tested this in the QA site over the past couple of days, and can confirm it all works as expected.  In the interest of getting this fix out to the production site as soon as possible, I'm going to merge into development.

vlagrassa:
Just for record-keeping: ack/nack codes are described at https://cloud.google.com/pubsub/docs/push

PR #388, opened 2023-08-10
Fix: Get Nemascan submission trait name from input file

vlagrassa:


src/pkg/caendr/caendr/services/tools/submit.py
```diff
@@ -461,6 +461,12 @@ def parse_data(cls, data, delimiter='\t'):
     # Compute data hash using entire file
     data_hash = get_file_hash(local_path, length=32)
 
+    # Open the file and extract the trait name from the header row
+    with open(local_path, 'r') as f:
+      csv_reader = csv.reader(f, delimiter=delimiter)
+      header_row = next(csv_reader)
+      data['trait'] = header_row[1]
```
r-vieira, line 468: Can we assert that the trait looks like a trait? Especially for mal-formed files.

r-vieira (approved):
Approved.

src/pkg/caendr/caendr/services/tools/submit.py
vlagrassa: We'd have to ask Erik if there's any expected format. As far as I can tell, a trait can be any arbitrary string (barring tabs & newlines), but we might(?) want to filter out special characters...

vlagrassa:
I've tested this in QA over the past few days, and can confirm it works.  In the interest of getting this fix out to the production site as soon as possible, I'm going to merge into development.

PR #389, opened 2023-08-11
Fix SVA table rebuild operation

vlagrassa:
Marking as draft until I can run a full SVA rebuild locally & confirm.  This is meant to address GCP issue, where rebuild op in CloudRun blew up the database (as in, increased memory until it became unusable).

- Make SVA table primary key a composite of species name & idx
  - Prevents key conflicts between species -- don't need to know anything about other species when rebuilding
- Better `SPECIES_LIST` parsing:
  -  Fix: recognizes empty string as empty (i.e. run on all species). Previously interpreted as empty list, and did nothing.
  - Validate incoming species names using `Species` class
- Improved logging:
  - List table names in more readable format when dropping/recreating
  - Show row counts before & after species-based drop operations
  - Track progress through batched generators


vlagrassa:
The SVA table bug ended up being a problem with the file, so these code changes don't fix it.  However, this branch still does improve the ETL pipeline.

While working with trait files for the phenotype db table, I found it useful to include the changes from this branch.  I'm opening this PR first, then the other PR, to pull the changes into `development` in smaller chunks.

src/modules/db_operations/main.py
```diff
@@ -79,14 +80,32 @@ def etl_operation_append_log(message = ""):
 db.init_app(app)
 
 
+def parse_species_list(species_list):
```
r-vieira, line 83: consider using default param value, eg: (species_list = None). Then check with `if species_list is None:`

src/pkg/caendr/caendr/services/sql/db.py
```diff
@@ -57,32 +57,51 @@ def drop_tables(app, db, species=None, tables=None):
...
     else:
-      logger.info(f'Dropping tables: ${tables}')
-      db.metadata.drop_all(bind=db.engine, checkfirst=True, tables=tables)
-      logger.info(f'Creating tables: ${tables}')
-      db.metadata.create_all(bind=db.engine, tables=tables)
+      _tables = [ t.__table__ for t in tables ]
+      table_names = 'tables [' + ', '.join([ t.__tablename__ for t in tables ]) + ']'
+
+      logger.info(f'Dropping {table_names}...')
+      db.metadata.drop_all(bind=db.engine, checkfirst=True, tables=_tables)
```
r-vieira, line 74: would this need to be changed to avoid #drop_all once we have phenotype database tables (metadata + data)?

r-vieira (approved):
Looks good. Let's discuss offline the #drop_tables() in the context of the upcoming phenotypeDB

PR #392, opened 2023-08-22
Added password guidelines and change button color when requirements a…

natalieroman:
Enhanced password reset UI to give user feedback on when password requirements are met. Updated template layout for Bootstrap 5.


r-vieira (approved):
looks good

PR #393, opened 2023-08-31
Re-enable new species for Heritability tool

vlagrassa:


src/modules/heritability_proxy/build-and-push-starter.sh
```diff
@@ -15,7 +15,7 @@ set +a
 source ../../../env/${ENV}/global.env
 
 # this requires reading the respective global.env above
-IMAGE_URI=gcr.io/${GOOGLE_CLOUD_PROJECT_ID}/heritability:v0.02-debug
+IMAGE_URI=gcr.io/${GOOGLE_CLOUD_PROJECT_ID}/heritability:v0.04
```
r-vieira, line 18: is there a way to avoid hardcoding version information in this file and instead load it from the global.env?

src/modules/heritability_proxy/run-gls-base.sh
```diff
@@ -9,7 +9,7 @@ fi
 echo "Env: ${ENV}"
 source ../../../env/${ENV}/global.env
 
-IMAGE_URI=gcr.io/${GOOGLE_CLOUD_PROJECT_ID}/heritability:v0.02-debug
+IMAGE_URI=gcr.io/${GOOGLE_CLOUD_PROJECT_ID}/heritability:v0.03-debug
```
r-vieira, line 12: similar to previous note -- can we avoid hardcoding version numbers here and load them from the global.env?

r-vieira (approved):
See if we can load the version info from the `global.env` instead of hardcoding it in these files.

src/modules/heritability_proxy/run-gls-base.sh
vlagrassa: Fixed in #394.

PR #394, opened 2023-09-07
Fix: Require login to access Heritability tool

vlagrassa:
A few miscellaneous touch-ups:
- Require user to login to see the main Heritability tool page. All other endpoints require login, so I think this is an oversight from the older architecture (which had a separate `create` endpoint). See discussion in commit message.
- Read container name & version from global `.env` files.  See discussion in #393.
- Update `launch.json` to create a local `.db` file in a `temp` directory, instead of inside the module folder.


r-vieira (approved):
approved

PR #396, opened 2023-09-27
Fix data release blueprints

vlagrassa:
- Merge the duplicate page at `https://caendr.org/data/` into `https://caendr.org/data/data-release`
- Move the `/data-release` prefix into the blueprint (so we're not repeating it in every endpoint)


r-vieira (approved):
looks good

PR #397, opened 2023-09-27
Pull CloudRun condition state enum from Google API

vlagrassa:
I was able to find an enum for the Condition State (succeeded, failed, etc) in the [`google-cloud-run`](https://cloud.google.com/run/docs/reference/rest/v2/Condition#State) library.

However, the Condition type is a `string` (not an `enum`) in the [docs](https://cloud.google.com/run/docs/reference/rest/v2/Condition), and I wasn't able to find a specification that lists all the possible values...


r-vieira (approved):
nice!

PR #398, opened 2023-09-27
Miscellaneous fixes (obsolete db close, API error codes, heritability report tweaks)

vlagrassa:
A few minor tweaks, bundled together:

- Update a few API error codes, to be more semantically accurate
- Remove obsolete code to explicitly close db on app context teardown
  - This is handled by Flask SQLAlchemy
  - I originally wrote these functions trying to debug the stale GCP connection issue (see #387 for the actual fix)
- Heritability report:
  - Update graph size on fullscreen
  - Order AssayNumber numerically (1, 2 ... 9, 10 ...) instead of alphabetically (1, 10, 11 ... 2, 3 ...)

### DIFF
--- a/src/modules/site-v2/templates/tools/heritability_calculator/result.html
+++ b/src/modules/site-v2/templates/tools/heritability_calculator/result.html
@@ -197,6 +197,7 @@
         return color;
     }
 
+    var renderPlot;
     var cpal = []
     var loadPlot = function(jj){
         df = [];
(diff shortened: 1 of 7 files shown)


src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -209,14 +210,26 @@
...
-        width = document.getElementById("svgchartarea").offsetWidth - 100; /* 100px is from the chart legend*/
-        height = 9 * width / 16;
-        console.log(width)
-        var chart = d3.exploding_boxplot(df, {y:'Value', group:'Strain', color:'AssayNumber', label:'TraitName', pal:cpal})
-                      .width(width)
-                      .height(height);
-        document.getElementById("svgchartarea").innerHTML = "";
-        chart('#svgchartarea');
+        // Create & save a function to compute the chart size & re-render it
+        renderPlot = function() {
```
r-vieira, line 214: when declaring JS vars in new code, prefer `let` and `const`. For lambda-type functions, use `const`

r-vieira: consider `const renderPlot = ...`

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -209,14 +210,26 @@
...
+          // Clear the current chart area
+          // Otherwise chart can't shrink - width computation will always be based on existing size
+          document.getElementById("svgchartarea").innerHTML = "";
+
+          // Take up as much horizontal space as available, keeping a 16:9 ratio
+          width = document.getElementById("svgchartarea").offsetWidth - 100; /* 100px is from the chart legend*/
+          height = 9 * width / 16;
+
+          // Create the chart and render in the page
+          var chart = d3.exploding_boxplot(df, {y:'Value', group:'Strain', color:'AssayNumber', label:'TraitName', pal:cpal})
```
r-vieira, line 225: prefer `const chart = ...`

r-vieira (approved):
looks good

src/modules/site-v2/templates/tools/heritability_calculator/result.html
vlagrassa: Fixed in 94bdca02fdd6f97a4d5ba2b37220d4be65abe2a8

src/modules/site-v2/templates/tools/heritability_calculator/result.html
vlagrassa: In this particular case, I'm actually re-assigning the value of `renderPlot` (declared on line 207) to make it available outside the inner function context.  (Not the best solution, but it involved the least code restructuring.)

For the top-level declaration of `renderPlot`, updated to `let` in 94bdca02fdd6f97a4d5ba2b37220d4be65abe2a8

PR #399, opened 2023-09-29
Enhancement/local file class

vlagrassa:
Quick utility class I put together a few weeks ago to make file uploads safer & to teach myself how to use context managers (`__enter__` and `__exit__`).

Lifecycle of a local file is now managed by the `LocalFile` class, meaning it will automatically be deleted off the server once the view function is done using it.  This pattern let me bundle the upload & delete logic together in a single file.

### DIFF
--- a/src/modules/site-v2/base/utils/local_file.py
+++ b/src/modules/site-v2/base/utils/local_file.py
@@ -1,0 +1,117 @@
+import os
+
+from werkzeug.utils import secure_filename
+
+from caendr.models.error import FileUploadError
+from caendr.utils.data import unique_id
+
+
+
+UPLOAD_DIR = os.path.join('./', 'uploads')
+os.makedirs(UPLOAD_DIR, exist_ok=True)
+
+
+class LocalFile():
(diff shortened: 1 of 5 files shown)


src/modules/site-v2/base/utils/local_file.py
```diff
@@ -0,0 +1,117 @@
...
+    '''
+      Ensure the file is deleted from local server storage.
+
+      Return:
+        removed (bool): Whether the file existed.
+    '''
+    try:
+      os.remove(self.__local_path)
+      return True
+    except FileNotFoundError:
```
r-vieira, line 84: we might have a few more OS related exceptions here (Eg: permissions/file locks/etc). Consider handling more edge cases.

r-vieira (approved):
very nice job on the inline comments!👍

src/modules/site-v2/base/utils/local_file.py
vlagrassa: Fixed in eb8a8f2c28e2f9edcfb4a62d1b4c0fe649601755

PR #400, opened 2023-09-29
Include narrow-sense heritability in report page

vlagrassa:
It looks like the heritability tool has always output this value, we just weren't reading & displaying it.  Old reports should retroactively display the value.

### DIFF
--- a/src/modules/site-v2/templates/tools/heritability_calculator/result.html
+++ b/src/modules/site-v2/templates/tools/heritability_calculator/result.html
@@ -52,9 +52,16 @@
       {% endif %}
       {% endif %}
     </p>
-    {%- if result %}
+    {%- if result and result['broad-sense'] %}
     <p>
-      <strong>Broad-Sense Heritability (H<sup>2</sup>):</strong> {{ (result['H2']*100)|round(2) }}% <small>(range {{ (result['ci_l']*100)|round(2) }}% to {{ (result['ci_r']*100)|round(2) }}%)</small>
+      <strong>Broad-Sense Heritability (H<sup>2</sup>):</strong> {{ (result['broad-sense']['H2']*100)|round(2) }}%
+      <small>(range {{ (result['broad-sense']['ci_l']*100)|round(2) }}% to {{ (result['broad-sense']['ci_r']*100)|round(2) }}%)</small>
+    </p>
+    {%- endif %}
+    {%- if result and result['narrow-sense'] %}
+    <p>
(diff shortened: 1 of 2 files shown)


src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -52,9 +52,16 @@
       {% endif %}
       {% endif %}
     </p>
-    {%- if result %}
+    {%- if result and result['broad-sense'] %}
```
r-vieira, line 55: do we need a default value in `broad-sense` is not set?

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -52,9 +52,16 @@
...
     <p>
-      <strong>Broad-Sense Heritability (H<sup>2</sup>):</strong> {{ (result['H2']*100)|round(2) }}% <small>(range {{ (result['ci_l']*100)|round(2) }}% to {{ (result['ci_r']*100)|round(2) }}%)</small>
+      <strong>Broad-Sense Heritability (H<sup>2</sup>):</strong> {{ (result['broad-sense']['H2']*100)|round(2) }}%
```
r-vieira, line 57: if applicable, consider creating a custom formatter for `*100)|round(2)`

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -151,7 +158,7 @@
             }
         }
 
-    var trait = "{{ trait }}"
+    var trait = "{{ hr.trait }}"
```
r-vieira, line 161: prefer `let` and `const` when declaring new JS vars

r-vieira (approved):
looks good

src/modules/site-v2/templates/tools/heritability_calculator/result.html
vlagrassa: Created in 9b53012d31b499518b785543eabb3e0cb09bc428

src/modules/site-v2/templates/tools/heritability_calculator/result.html
vlagrassa: Fixed in 94bdca02fdd6f97a4d5ba2b37220d4be65abe2a8 (actually removed this variable)

src/modules/site-v2/templates/tools/heritability_calculator/result.html
vlagrassa: Might be worth asking Erik? This should only be possible if there was an error running the tool (in which case the whole report would be an error) or if something went wrong in the storage / retrieval...

PR #401, opened 2023-10-05
Minor fixes for recent PRs

vlagrassa:
Addresses comments in #398, #399, and #400


src/modules/site-v2/application.py
```diff
@@ -294,6 +294,10 @@ def _jinja2_filter_species_italic(text):
     text = text.replace('C. elegans', '<i>C. elegans</i>')
     return text
 
+  @app.template_filter('as_percent')
```
r-vieira: consider `percent` instead of `as_percent`? Trying align with existing filter names.

src/modules/site-v2/base/utils/local_file.py
```diff
@@ -83,6 +84,9 @@ def remove(self):
       return True
     except FileNotFoundError:
       return False
+    except Exception as e:
```
r-vieira, line 102: We might need to expand some of the exceptions, e.g.: file permissions/file locks/etc, before returning `True`.

r-vieira (changes requested):
minor question on naming and handling file removal.

src/modules/site-v2/application.py
vlagrassa: Fixed in [8e2842c](https://github.com/AndersenLab/CAENDR/pull/401/commits/8e2842c2c40c21106e705ed32eed5fc0ab612a07)

src/modules/site-v2/base/utils/local_file.py
vlagrassa: What do you think of [fd4192d](https://github.com/AndersenLab/CAENDR/pull/401/commits/fd4192d11f84f970839ea6e4653b115131cf7870)?

In the interest of not spending too much time on this -- I check if the file still exists, and suppress the error if it doesn't; otherwise, I propagate the error to be handled elsewhere.  It's not a watertight solution, but it should prevent a failed `remove` from flying under the radar...

r-vieira: sounds good.

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -196,17 +196,17 @@
     }
 
     function getRandomColor() {
-        var letters = '123456789ABCDEF'.split('');
-        var color = '#';
-        for (var i = 0; i < 6; i++ ) {
+        const letters = '123456789ABCDEF'.split('');
```
r-vieira, line 199: Do we need to worry about contrast against background/other elements on the page? If so, we might need to restrict the range of the output color based on some constraints.

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -229,58 +229,56 @@
...
         }
 
         // Call the new render plot function
         renderPlot();
     }
 
-    async function drawImg(canvas, image, DOMURL, w, h){
+    async function drawImg(canvas, image, DOMURL, w, h) {
         return new Promise(resolve => {
           image.onload = function () {
```
r-vieira, line 244: do we need to fail gracefully when image.onerror happens => reject()?

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -322,22 +320,22 @@
...
         doc.fromHTML('of strains. Generally, it is the ratio of genetic variance to total (genetic plus environmental) variance.', 20, 100, {'align': 'justify', 'maxWidth': doc.internal.pageSize.getWidth()-40},function(){},margins)
 
-        var a = document.createElement('a');
-            var pdfblob = new Blob([ doc.output('blob') ], { type : 'application/pdf'});
-            a.href = window.URL.createObjectURL(pdfblob);
-            a.download = "{{ fname }}.pdf";
-            a.click();
-        })
-        image.src = url;
+        const a = document.createElement('a');
```
r-vieira, line 323: we could bundle this logic as a `download_file()` function.

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -322,22 +320,22 @@
...
 
-    var getPDF = function(){
+    function getPDF() {
       loadCanvas();
     }
 
-    $(document).ready(function(){
-      if (chartJson.length != 0){
-        var node= document.getElementById("svgchartarea");
+    $(document).ready(function() {
```
r-vieira, line 336: unless we rely on jQuery, we can use standards: window.addEventListener('load', () => {}, false);

r-vieira (approved):
looks good! see inline comments.

src/modules/site-v2/templates/tools/heritability_calculator/result.html
vlagrassa: Good question -- that might be a question for Natalie.

PR #402, opened 2023-10-06
Displaying file upload strain errors all at once

vlagrassa:
- Validators are now split into two functions: the `step` and the `final`
  - For `validate_strains`, the `step` keeps a list of invalid lines, categorized by error type: wrong species, unknown strain, blank strain, and (optionally) duplicate strain
  - For `validate_strains`, the `final` runs through each category, and displays the first one that has one or more errors
  - For the other validators, only the `step` is required (remains unchanged)
- Error messages are meant to be human-readable
  - If 1-3 strains are invalid, display those in an inline list (using correct pluralization, commas + "and", etc)
  - If 4+ are invalid, include a dropdown (a BS `collapse` element) with the full list
  - Species names are italicized
- Submission API calls accept an optional URL variable `reloadonerr`
  - If `0`, returns the error message without using Flask `flash`. The JS function will manually show the error without reloading the page
  - If `1` or unset, uses `flash` to display the message. The JS function will reload the page to make this appear

### DIFF
--- a/src/modules/site-v2/base/utils/tools.py
+++ b/src/modules/site-v2/base/utils/tools.py
@@ -168,7 +168,11 @@ def try_submit(EntityClass, user, data, no_cache):
       msg += f' (Line: { ex.line })'
 
     # Return the error message
-    return { 'message': msg }, 400
+    return {
+      'message': msg,
+      'full_msg_link': ex.full_msg_link,
+      'full_msg_body': ex.full_msg_body,
+    }, 400
 
   # General error
   except Exception as ex:
--- a/src/modules/site-v2/base/views/tools/genetic_mapping.py
+++ b/src/modules/site-v2/base/views/tools/genetic_mapping.py
@@ -116,7 +116,7 @@ def submit():
       response, code = try_submit(NemascanMapping, user, data, no_cache)
 
       # If there was an error, flash it
-      if code != 200 and int(request.args.get('reloadonerr', 1)):
+      if code != 200 and int(request.args.get('reloadonerror', 1)):
         flash(response['message'], 'danger')
 
       # If the response contains a caching message, flash it
--- a/src/modules/site-v2/base/views/tools/genetic_mapping.py
+++ b/src/modules/site-v2/base/views/tools/genetic_mapping.py
@@ -123,10 +123,10 @@ def submit():
     response, code = try_submit(NemascanMapping, user, data, no_cache)
 
     # If there was an error, flash it
-    if not code == 200:
+    if code != 200 and int(request.args.get('reloadonerr', 1)):
       flash(response['message'], 'danger')
 
-    elif response.get('message') and response['ready']:
+    elif response.get('message') and response.get('ready', False):
       flash(response.get('message'), 'success')
 
     # Return the response
--- a/src/modules/site-v2/templates/_scripts/submit-job.js
+++ b/src/modules/site-v2/templates/_scripts/submit-job.js
@@ -18,7 +18,24 @@ $.ajaxSetup({
   If defining with `as_form_data` = True, must include `ajax_setup` on the page (defined above).
 #}
 {% macro def_submit_job(tool_name, as_form_data=false, func_name='submit_job') %}
-function {{func_name}}(data, modal_id, new_tab=false) {
+function {{func_name}}(data, modal_id, new_tab=false, propagate_err=true) {
+
+  // Gather URL variable(s)
+  let url_vars = [];
+  if (propagate_err) {
+    url_vars.push('reloadonerr=0');
+  }
+  {%- if session["is_admin"] %}
+  if ($('#no_cache_checkbox').prop('checked')) {
+    url_vars.push('nocache=1');
+  }
+  {%- endif %}
+
+  // Create full URL, appending URL vars if any were created
+  let url = "{{ url_for(tool_name + '.submit') }}";
+  if (url_vars.length) {
+    url += '?' + url_vars.join('&');
+  }
 
   return $.ajax({
     type: "POST",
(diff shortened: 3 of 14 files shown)


src/modules/site-v2/base/utils/tools.py
```diff
@@ -122,7 +122,11 @@ def try_submit(EntityClass, user, data, no_cache):
       msg += f' (Line: { ex.line })'
 
     # Return the error message
-    return { 'message': msg }, 400
+    return {
+      'message': msg,
+      'full_msg_link': ex.full_msg_link,
+      'full_msg_body': ex.full_msg_body,
+    }, 400
```
r-vieira, line 129: `400` sounds like a good candidate for an `enum/class`. I.e.: `HttpStatusCode.BAD_REQUEST`

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -115,10 +115,12 @@ def submit():
       # Try submitting the job & returning a JSON status message
       response, code = try_submit(NemascanMapping, user, data, no_cache)
 
-      # If there is an error or a ready message, flash it
-      if not code == 200:
+      # If there was an error, flash it
+      if code != 200 and int(request.args.get('reloadonerr', 1)):
```
r-vieira: consider `reloadonerror` instead of shorter version.

r-vieira: `200` => HttpStatusCode.OK enum/class?

src/modules/site-v2/templates/_scripts/submit-job.js
```diff
@@ -18,20 +18,32 @@ $.ajaxSetup({
   If defining with `as_form_data` = True, must include `ajax_setup` on the page (defined above).
 #}
 {% macro def_submit_job(tool_name, as_form_data=false, func_name='submit_job') %}
-function {{func_name}}(data, modal_id, new_tab=false) {
+function {{func_name}}(data, modal_id, new_tab=false, propagate_err=true) {
```
r-vieira: prefer `propagate_error` for clarity over abbreviations.

r-vieira: I appreciate the `c-macro-ish-ness` nature of `{{ func_name }}`. Is there a chance it might be null/undefined/use reserved keyword, etc?

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -98,3 +98,54 @@ function force_download(e) {
...
+  // Inserting as text protects against code injection
+  const node = create_node(raw_html);
+  node.firstElementChild.innerText = message;
+
+  // If both full message fields are provided, add as a link & collapse dropdown
+  if (full_msg_link && full_msg_body) {
+    node.firstElementChild.innerText += ' ';
+
+    // Create a unique ID to link the trigger & collapse elements
+    const id = `err-dropdown-${(new Date()).getTime()}`;
```
r-vieira: prefer `error` vs `err`

src/modules/site-v2/templates/tools/pairwise_indel_finder/submit.html
```diff
@@ -549,7 +549,7 @@ <h1 class="modal-title fs-5" id="generate-primers-modal-label">Thank You</h1>
   }
 
   // Submit the job, opening results in a new tab if applicable
-  submit_job(data, '#generate-primers-modal', true)
+  submit_job(data, '#generate-primers-modal', true, false)
```
r-vieira: consider using a config object instead of sequencial booleans, (...,..., { foo: true, bar: false}) for clarity of order.

r-vieira (changes requested):
review inline comments and lmk what you think.

src/modules/site-v2/base/utils/tools.py
vlagrassa: Happy to make the enum if you think it would help...  I'm tempted to say the code numbers are pretty transparent, but I can see how they might be confusing, esp. when they're returned a few function calls down (as opposed to from the view function directly)

src/modules/site-v2/templates/_scripts/utils.js
vlagrassa: Fixed in [b137d57](https://github.com/AndersenLab/CAENDR/pull/402/commits/b137d5709aff41d74bc8177241ca161ae2e3d02d)

src/modules/site-v2/templates/_scripts/submit-job.js
vlagrassa: Fixed the first part in [b137d57](https://github.com/AndersenLab/CAENDR/pull/402/commits/b137d5709aff41d74bc8177241ca161ae2e3d02d)

src/modules/site-v2/base/views/tools/genetic_mapping.py
vlagrassa: Fixed the first part in [b137d57](https://github.com/AndersenLab/CAENDR/pull/402/commits/b137d5709aff41d74bc8177241ca161ae2e3d02d)

src/modules/site-v2/templates/tools/pairwise_indel_finder/submit.html
vlagrassa: Fixed in [27fe543](https://github.com/AndersenLab/CAENDR/pull/402/commits/27fe543f3e581ff847e79c2b5470b48b31e2d903)

src/modules/site-v2/templates/_scripts/submit-job.js
vlagrassa: To the second point - as it turns out, none of the calling sites use the `func_name` argument to change the name.  I think I put this in so we'd have a way to manually avoid name collisions (rather than the parameter being filled in by a variable), but that ended up not being necessary...

(To clarify -- I was envisioning something like, the following, using a string:
`{{ def_submit_job( ..., func_name='submit_job_foobar' ) }}`

...rather than something using a variable:
`{{ def_submit_job( ..., func_name=foobar ) }}`

...but since we don't use either, I could also just remove the parameter)

src/modules/site-v2/base/utils/tools.py
r-vieira: I see -- maybe I'm looking at this from the perspective of new developer approaching the codebase where it might be useful to know the domain of the 'magic numbers' in code, eg: is the number referring to some internal status code of the application layer?Http? others? in this case, it helps to clarify that we are setting the HttpStatusCode semantically in the use of enums (of course, at the expense of extra verbosity 😬).

src/modules/site-v2/templates/_scripts/submit-job.js
r-vieira: got it, thanks!

src/modules/site-v2/base/utils/tools.py
vlagrassa: Makes sense. I'll add a card for it in BC

PR #403, opened 2023-10-13
Enhancement/entity class abstraction

orzut:
Moved the 'soft delete' functionality of 'Cart' to separate subclass 'DeletableEntity'


src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -35,29 +32,30 @@ def get_props_set(cls):
...
-
-    # Try finding and returning unique card for this user
+    
+    # Try finding cart(s) for user
     try:
-      return Cart.query_ds_not_deleted('user', email, required=True)
+      # Run the query and try to find not deleted cart(s) for the user
+      carts = cls.query_ds_not_deleted('user', email, required=True)
 
+      if len(carts) == 1:
```
vlagrassa: A little trick for this part -- if you walk through the logic, the `else` branch will actually work in both cases!  (Since sorting a single-element list doesn't do anything to it.)  If you collapse this `if-else` statement into a single line (just line 53), you end up with a single line of code and with fewer edge cases.

If you decide to go this route, you'd probably want to adjust the comment to make it clear what you're doing.

vlagrassa (commented):
Looks great! The comments are very helpful :)

src/pkg/caendr/caendr/models/datastore/cart.py
orzut: Good catch! Removed if/else condition in the commit [b066433](https://github.com/AndersenLab/CAENDR/pull/403/commits/b066433938e167756db66dc9c06f9bb7a1b330cd)

vlagrassa (approved):

src/pkg/caendr/caendr/models/datastore/deletable_entity.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # If no matching entities found, return None
+    if len(matches) == 0:
+      if required:
+        raise NotFoundError(cls.kind, {key: val})
+      else:
+        return None
+
+    # If looking up for unique entity and multiple were found, raise an error
+    elif len(matches) > 1 and unique:
```
r-vieira, line 58: minor style detail on `if` statements -- prefer early guards/departures as part of control flow. For instance, consider: 

```js
# If no matching entities found, return None
if len(matches) == 0:
    raise NotFoundError(cls.kind, {key: val}) if required else return None

# If looking up for unique entity and multiple were found, raise an error
if len(matches) > 1 and unique:
  raise NonUniqueEntity( cls.kind, key, val, matches )

# If one or more entities were found and 'unique' is set to False, return all of them
return matches
```

src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -35,29 +32,27 @@ def get_props_set(cls):
...
+      carts = cls.query_ds_not_deleted('user', email, required=True)
 
+      # If one cart was found, return it
+      # Or if multiple carts were found, sort by date modified on and return latest
+      return cls.sort_by_modified_date(carts, reverse=True)[0]
 
-    # User doesn't have a cart
+    # If a user doesn't have a cart, create one
     except NotFoundError:
       return Cart(**{'user': email, 'items': [], 'version': 0})
```
r-vieira: tl;dr;not part of the PR, but consider simplifying the complexity of the function `lookup_by_user()` by abstracting the cart creation. In other words, this function currently *needs* to understand *how* to create a new cart (e.g.: the items must be an empty array, and the version must be `0`). Consider abstracting the cart creation to a new function or factory method that initializes all the values. This approach centralizes the logic behind the cart creation, and avoids having to update all the places in the code that create Carts, if the cart creation logic is updated at a later time.

Also, if it makes sense, consider setting default values at the class level during initialization (e.g. items:[], and version:0) instead of keep track of that logic on the calling site.

r-vieira (changes requested):
Looks great! Please review minor inline comments.

src/pkg/caendr/caendr/models/datastore/deletable_entity.py
orzut: fixed in the commit [5d4f4af](https://github.com/AndersenLab/CAENDR/pull/403/commits/5d4f4affda4c0bf046d2d82fc16f0f96feb95d70)

src/pkg/caendr/caendr/models/datastore/cart.py
orzut: Please see commit [5a94a1d](https://github.com/AndersenLab/CAENDR/pull/403/commits/5a94a1d32af4ba669e0a0893f1754f7018ea0b1c) for changes and let me know if this is what you're looking for.

r-vieira (approved):
looks good

PR #405, opened 2023-10-20
Move scatter histograms code to JS function

vlagrassa:
Turns the D3 scatter histograms plot code into a set of JS functions


orzut:
Looks good! Thanks for organizing everything in a nicer way!

orzut (approved):

PR #406, opened 2023-11-03
Fixing bug with not being able to add multiple strain sets

orzut:
- checking for species name when adding items to cart (same with removing from the cart)
- displaying species name for strain sets in Confirmation page

### DIFF
--- a/src/modules/site-v2/base/views/strains.py
+++ b/src/modules/site-v2/base/views/strains.py
@@ -225,7 +225,7 @@ def order_page_post():
           return redirect(url_for('request_strains.order_page_index'))
         
         if form.shipping_service.data == 'Flat Rate Shipping':
-          cartItems.append({'name': 'Flat Rate Shipping'})
+          cartItems.append({'name': 'Flat Rate Shipping', 'species': ''})
         for item in cartItems:
           item_price = Cart.get_price(item)
           item['price'] = item_price
@@ -237,7 +237,7 @@ def order_page_post():
                       'is_donation': False}
         order_obj.update(form.data)
         order_obj['phone'] = order_obj['phone'].strip("+")
-        order_obj['items'] = '\n'.join(sorted([f"{item['name']}:{item['price']}" for item in cartItems]))
+        order_obj['items'] = '\n'.join(sorted([f"name: {item['name']}, species: {item['species']}, price: {item['price']}" for item in cartItems]))
         order_obj['invoice_hash'] = str(uuid.uuid4()).split("-")[0]
         order_obj["order_confirmation_link"] = url_for('request_strains.order_confirmation', invoice_hash=order_obj['invoice_hash'], _external=True)
         send_email({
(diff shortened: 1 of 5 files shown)


src/modules/site-v2/templates/order/order.html
```diff
@@ -57,11 +57,11 @@
                 {% for item in cartItems %}
                 <tr class='cartItem'>
                     <td>{{ loop.index }}</td>
-                    <td><em>{{ item.species }}</em></td>
+                    <td><em>{{ item.species|capitalize|replace('_', '. ') }}</em></td>
```
vlagrassa, line 60: It's a bit hidden, but `item.species.short_name` should do this same operation (see line ~56 in `models/datastore/species.py`).  I don't think we'll ever need to change how the "short name" works, but if we did, it's good to have it defined in just one place.

src/modules/site-v2/templates/order/order_confirm.html
```diff
@@ -52,7 +52,7 @@ <h3>Order Information</h3>
                 {% for item in items %}
                 <tr>
                     <td>{{ loop.index }}</td>
-                    <td>{{ item['strain'] }}</td>
+                    <td>{{ item['name'] }}</td>
                     <td><em>{{ item['species']|capitalize|replace('_', '. ') }}</em></td>
```
vlagrassa, line 56: Same comment here as above - you can use the species `short_name` property to generate this value

src/pkg/caendr/caendr/models/datastore/cart.py
vlagrassa, line 1: Quick thought on this --

It might be a nice improvement to add a "helper" class called `CartItem` (or something similar), which defines what fields an item in the cart should have.  You could then use that class to check equality (by defining the `__eq__` method).

I don't think you need to do that right now, but it might be something to write down for later -- I have a number of cards in Basecamp with smaller improvements I'd like to make when I have the time

src/modules/site-v2/base/views/strains.py
```diff
@@ -225,7 +225,7 @@ def order_page_post():
           return redirect(url_for('request_strains.order_page_index'))
         
         if form.shipping_service.data == 'Flat Rate Shipping':
-          cartItems.append({'name': 'Flat Rate Shipping'})
+          cartItems.append({'name': 'Flat Rate Shipping', 'species': ''})
```
vlagrassa, line 228: Can/should this use `Cart.add_item`, instead of appending directly?

This is a bit subjective, but I think it's usually safer to ask the `Cart` object for information, rather than using the `items` prop directly.  That way, if we need to do something fancy behind-the-scenes with `Cart`, we don't have to change any of this code that uses it.

--

Not required for this PR, but a few ideas I'll mention here anyways:

- If you want to learn more about Python "dunder" methods (dumb name I know), you could try implementing `Cart.__len__` and using `len(users_cart)` a few lines above here (instead of calling `len` on `users_cart['items']`.
- You might also consider making a `Cart.is_empty()` method, in the interest of hiding implementation details inside the `Cart` class.

You could try looking into those now, or put them off for later if you have more urgent work :)

src/modules/site-v2/base/views/strains.py
```diff
@@ -341,21 +341,18 @@ def order_confirmation(invoice_hash):
...
-  strains = get_strains()
-  for item in items:
-    for strain in strains:
-      if item['strain'] == strain.isotype:
-        item['species'] = strain.species_name
-        break
-      else:
-        item['species'] = ''
+  # Parse the individual items in the order into a list of dicts
+  order_obj["items"] = [ x for x in order_obj['items'].split("\n") ]
```
vlagrassa, line 345: This is a bit pedantic, but since `str.split` outputs a list, you don't need the `[ x for x in ... ]` list comprehension here.

(Sometimes you might see `x for x in ...` if there's an `if` clause after it -- that's how you'd write a filter using list comprehensions.  Alternatively, if all you want to do is convert to a list, you can use the built-in function `list()`.  In this particular case, I think you're safe to use `split` directly.)

src/modules/site-v2/templates/order/order.html
```diff
@@ -232,7 +232,7 @@ <h5 class="modal-title" id="deleteModalTitle">Are you sure you want to delete th
 
     $('#deleteItem').click(function(e){
         e.preventDefault()
-        const data = {csrf_token: $('#csrf_token').val(), itemToRemove: $(this).attr('info')}
+        const data = {csrf_token: $('#csrf_token').val(), itemToRemove: {name: $(this).data('value').name, species: $(this).data('value').species}}
```
vlagrassa, line 235: I had no idea you could access HTML `data` fields as objects like this. Very cool! :)

vlagrassa (changes requested):
Nice!  I left a few comments for cleaning up the code a bit, but overall I think this looks really good :)

My only real comment is to try and use `Cart` methods more directly -- e.g. using `Cart.add_item` instead of `Cart['items'].append`.  That lets you control the behavior for adding items *within* the `Cart` class, without having to worry about it from the calling function.

I also left a few ideas for future improvements.  Those don't have to be implemented for this PR, but it might be worth keeping track of them for later.

Thanks!

PR #409, opened 2023-11-14
Minor changes

orzut:
Minor changes for the previous PR. 
Note: getting `short_name` property for species in the Order pages wasn't straight forward, because species values in the cart are stored as string (they are not linked to Species class). Let me know if the approach I took makes sense.


src/modules/site-v2/base/views/strains.py
```diff
@@ -342,16 +346,20 @@ def order_confirmation(invoice_hash):
...
-  for row in order_obj['items']:
+  for row in order_obj['items'].split("\n"):
     arr = row.split(', ')
     item_dict = {}
     for x in arr:
       k, v = x.split(': ')
       if k == 'price':
         v = float(v)
       item_dict[k] = v
+    species = item_dict.get('species')
```
vlagrassa, line 352: Aha, I like the idea here!  This is definitely a valid solution, and I don't want to fixate on smaller details when we have so much else to work on.

That said -- I wonder if it's possible to add the `Species` object to the `item_dict` directly? e.g. something like:
```
item_dict['species'] = Species.from_name(species) if species else None
```
Then the full `Species` object would be available in the template.

Might not be worth doing now, but I'd at least hold onto the idea as an improvement for later.  It might make sense to turn the cart items into their own `class` at some point, and if so, you could make this change then.

src/pkg/caendr/caendr/models/datastore/cart.py
```diff
@@ -26,6 +26,10 @@ def __init__(self, name_or_obj = None, *args, **kwargs): # If nothing passed for
       self['items'] = []
     if not self['version']:
       self['version'] = 0
+
+  # Get the number of items in the cart
+  def __len__(self):
```
vlagrassa, line 31: Nice!

src/modules/site-v2/base/views/strains.py
```diff
@@ -303,7 +303,7 @@ def order_page_index():
   flash(Markup("<strong>Please note:</strong> although the site is currently able to accept orders, orders will <u>not ship</u> until Fall 2023."), category="warning")
 
   if not user and not cart_id:
-    cartItems = []
+    users_cart = []
```
vlagrassa: Hmm... I see what you're doing here, setting `users_cart` to an object with `len == 0` so it triggers the check below (line 316).

If possible, though, it's good practice to keep all uses of a variable the same type, for code clarity.  In this case, `users_cart` is assigned a `list` in the first branch of the `if` statement, and a `Cart` in the second branch, which makes it harder to reason about in later lines of code.

(In your specific case, you nullify the difference immediately, so this solution works out.  The problem is, if you (or someone else) has to refactor the code later on, this might turn into a subtle bug.)

Is it possible to create a new, empty `Cart` object here?  And if so, would that have length `0`?  Or do you need to have a user ID before you can make a cart?

(If this involves too much refactoring, feel free to mark it as a TODO and come back to it later.)

vlagrassa (commented):
Looks good!  I especially like how `len` & `add_item` turned out.

I added a few comments -- feel free to set these aside as TODOs for later, when we have more time to set aside for code structure & clarity.

src/modules/site-v2/base/views/strains.py
orzut: I like the idea, I'll create a BC card for this and put it to later ToDos

src/modules/site-v2/base/views/strains.py
orzut: I went with an early return here. Please see the changes in the commit 41f01790e9b01d56771328fb080c91592ba1618b and let me know what you think

vlagrassa:
Looks good to me! :)

vlagrassa (approved):

r-vieira (approved):
approved!

PR #410, opened 2023-11-20
Make file column validators into classes

vlagrassa:
Cleans up the validation logic.
- The `step` and `final` functions are now class methods instead of optional returns on an arbitrary dict, which standardized the logic for running the validators
- The column header validation is moved into the new `ColumnValidator` objects, which cleans up the syntax for `validate_file` (now takes a list of `ColumnValidators` rather than a list of semi-arbitrary dicts)


r-vieira (approved):
approved

PR #411, opened 2023-11-20
Report abstract base class, generic Runner class

vlagrassa:
Highlights:
- Replaces `DataJobEntity` class with `ReportEntity`, which subclasses from `Report` for a more generic interface and from `Entity` / `JobEntity` to fill out the abstract methods
- Removes job-specific subclasses of the `Runner` class, instead using the more generic implementation-based class (in our case, `GCPCloudRunRunner`)
- Moves input/output parsing & defining the run params to the job-specific `JobPipeline` subclasses
- Uses general `generate_blob_uri` function and `BlobURISchema` enum to create URLs, bucket & path tuples, etc for blobs in a standardized way

More specific comments inline


src/modules/site-v2/base/utils/tools.py
```diff
@@ -79,19 +79,14 @@ def try_submit(kind, user, data, no_cache):
     job = get_pipeline_class(kind=kind).create(user, data, no_cache=no_cache, valid_file_extensions=TOOL_INPUT_DATA_VALID_FILE_EXTENSIONS)
 
     # Schedule the job, if applicable
-    #   CachedDataError    -> the job already has results
+    #   JobAlreadyScheduledError -> a job computing these results has already been scheduled
```
vlagrassa, line 82: Part of the new explicit separation between job creation & submission. This error name is more accurate in the new flow, and is a bit more general (catches a job that's already been submitted & redirects to that ongoing submission)

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -107,10 +107,10 @@ def submit():
 
   # Upload input file to server temporarily, and start the job
   try:
-    with LocalFile(request.files.get('file'), valid_file_extensions=TOOL_INPUT_DATA_VALID_FILE_EXTENSIONS) as filepath:
+    with LocalFile(request.files.get('file'), valid_file_extensions=TOOL_INPUT_DATA_VALID_FILE_EXTENSIONS) as file:
```
vlagrassa, line 110: Now passing the `LocalFile` object to the submission code directly, which helps distinguish between:
- a string that should be uploaded as a file
- a local file object that should be uploaded as-is

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -209,10 +197,9 @@ def report(id):
 
     'id': id,
 
-    # URLs
-    'report_url': report_url,
-    'report_status_url': url_for("genetic_mapping.report_status", id=id),
-    'data_download_url': data_download_url,
+    # Links to the input data file and report output files, if they exist
```
vlagrassa, line 200: The new `Report` class tracks where input/output are stored, so we can use it to get the relevant filepaths (and to check if files exist at those paths)

src/modules/site-v2/base/views/tools/genetic_mapping.py
```diff
@@ -233,11 +220,11 @@ def report_fullscreen(id):
     abort(ex.code)
 
   # Download the report files, if they exist
-  if job.report.report_path is not None:
-    blob = get_blob(job.report.get_bucket_name(), job.report.report_path)
-    report_contents = blob.download_as_text()
-  else:
-    report_contents = None
+  report_contents = job.fetch_output()
```
vlagrassa, line 223: Similar to above, `Report` class manages access to input & output files

src/modules/site-v2/base/views/tools/heritability_calculator.py
```diff
@@ -165,10 +164,10 @@ def submit():
 
   # Upload input file to server temporarily, and start the job
   try:
-    with LocalFile(request.files.get('file'), valid_file_extensions=TOOL_INPUT_DATA_VALID_FILE_EXTENSIONS) as filepath:
+    with LocalFile(request.files.get('file'), valid_file_extensions=TOOL_INPUT_DATA_VALID_FILE_EXTENSIONS) as local_file:
```
vlagrassa, line 167: Passing `LocalFile` object to submission code (see comment above)

src/modules/site-v2/base/views/tools/heritability_calculator.py
```diff
@@ -276,18 +274,14 @@ def report(id):
 
     'ready': ready,
 
-    # TODO: The HTML file expects a variable called "fnam" -- is that a typo?
     'fname': datetime.today().strftime('%Y%m%d.') + trait,
 
     'hr': job.report,
     'data': data,
     'result': result,
+    'error': job.get_error(),
```
vlagrassa, line 282: `JobPipeline` object now provides access to the error, so we don't have to use the `PipelineOperation` (can instead let it be handled implicitly by the `Runner`)

src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
```diff
@@ -304,29 +303,26 @@ def report(id, file_ext=None):
       logger.error(f'Invalid interval "{data["site"]}" for Indel Finder report {id}')
       indel_start, indel_stop = None, None
 
+    # Extract the dataframe from the results
+    dataframe = result.get('dataframe', None)
 
-    # Update the result object with computed fields and generate a format table
-    # If result is None or empty, does nothing
-    result, format_table = modify_indel_primer_result(result)
```
vlagrassa, line 310: Instead of the somewhat ad-hoc `modify_indel_primer_result` function, this processing is now part of the `IndelPrimerJobPipeline` class's fetching logic

src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
```diff
@@ -340,7 +336,7 @@ def report(id, file_ext=None):
       'id': id,
 
       # Job status
-      'empty': result is None,
+      'empty': result['empty'],
```
vlagrassa, line 339: Whether or not the result is empty is passed back as an explicit dict field

src/modules/site-v2/templates/tools/heritability_calculator/result.html
```diff
@@ -48,7 +48,7 @@
     
       {% if hr.status == JobStatus.ERROR %}
       {% if session["is_admin"] %}
-      <p><strong>Error:</strong> <span class="text-danger">{{ operation.error }}</span></p>
+      <p><strong>Error:</strong> <span class="text-danger">(code {{ error.code }}) {{ error.message }}</span></p>
```
vlagrassa: The `JobPipeline` returns the error with a code & message, so this displays them more neatly

src/pkg/caendr/caendr/api/strain.py
```diff
@@ -170,14 +170,9 @@ def get_bam_bai_download_link(species, strain_name, ext, signed=False):
...
-  bam_prefix = BAM_BAI_PREFIX.get_string(SPECIES=species.name)
+  bam_prefix  = BAM_BAI_PREFIX.get_string(SPECIES=species.name)
 
-  blob_name = f'{bam_prefix}/{strain_name}.{ext}'
-
-  if signed:
-    return generate_download_signed_url_v4(bucket_name, blob_name)
-  else:
-    return generate_blob_url(bucket_name, blob_name)
+  return generate_blob_uri( bucket_name, bam_prefix, f'{strain_name}.{ext}', BlobURISchema.sign(signed) )
```
vlagrassa, line 175: The new unified `generate_blob_uri` function can handle optionally signing a URL, so no need to swap between two different functions here

(Specifically, `BlobURISchema.sign( ... )` will specify a signed URL if its argument is `True`, and an unsigned URL if `False`)

src/pkg/caendr/caendr/api/strain.py
```diff
@@ -228,22 +228,20 @@ def generate_bam_bai_download_script(species, release, signed=False):
...
+    bai_fname = f'{strain}.bam.bai'
 
     # Generate download URLs
-    if signed:
-      bam_url = generate_download_signed_url_v4(bucket_name, bam_path, expiration=expiration, credentials=credentials)
-      bai_url = generate_download_signed_url_v4(bucket_name, bai_path, expiration=expiration, credentials=credentials)
-    else:
-      bam_url = generate_blob_url(bucket_name, bam_path)
-      bai_url = generate_blob_url(bucket_name, bai_path)
+    bam_url = generate_blob_uri(bucket_name, bam_prefix, bam_fname, **sign_dict)
```
vlagrassa, line 237: New unified `generate_blob_uri` functions takes the path as a variadic list argument, so we can insert the `bam_prefix` here and let the function handle joining the paths (rather than assuming a `/` format here)

vlagrassa: It also handles signing based on the `schema`, which is passed in as part of `sign_dict`

src/pkg/caendr/caendr/models/datastore/data_job_entity.py
vlagrassa, line 1: This has been replaced by:
- `ReportEntity`: implements the abstract methods of `Report` using `Entity` and its subclasses
- `HashableEntity`: adds and manages a data hash prop to an entity

src/pkg/caendr/caendr/models/datastore/database_operation.py
```diff
@@ -62,32 +100,22 @@ def get_props_set(cls):
...
-  ## Special Properties ##
-
+  #
+  # Database Operation prop
+  #
 
   @property
   def db_operation(self):
-    return self.__db_operation
+    return self._get_enum_prop(DbOp, 'db_operation', None)
```
vlagrassa, line 109: Implemented & standardized `Enum` props using `Entity._set_enum_prop` and `Entity._get_enum_prop`

src/pkg/caendr/caendr/models/datastore/dataset_release.py
```diff
@@ -217,7 +217,7 @@ def check_fasta_file_exists(self):
   @staticmethod
   def get_fasta_filepath_url_template():
     obj = DatasetRelease.get_fasta_filepath_obj_template()
-    return TokenizedString( generate_blob_url(obj['bucket'], f'{ obj["path"] }/{ obj["name"] }{ obj["ext"] }') )
+    return TokenizedString.apply( generate_blob_uri, obj['bucket'], obj['path'], obj['name'] + obj['ext'], schema=BlobURISchema.HTTPS )
```
vlagrassa, line 220: The `apply` method applies a given function (`generate_blob_uri` here) to a set of `TokenizedString` arguments, handling the tokenization implicitly.

Side note: I originally designed the `TokenizedString` class to be able to carry tokens around, but now that I've been using it for a bit, I've found that there's almost never a practical use case for that (you pretty much always know all the tokens you want to use at the same time).  I don't want to focus too much energy on it now, since it does work, but I think a more streamlined class that just validates token replacements would be (a) equally useful and (b) less awkward.

src/pkg/caendr/caendr/models/datastore/entity.py
```diff
@@ -310,6 +316,36 @@ def modified_on(self):
...
 
 
+  def _get_raw_prop(self, key, fallback=None):
+    return self.__dict__.get(key, fallback)
+
+  def _set_raw_prop(self, key, val):
+    self.__dict__[key] = val
+
+
+  def _get_enum_prop(self, EnumType, key, fallback=None):
```
vlagrassa, line 326: Implements & standardizes Entities having `Enum` props.  I'd like to redesign the entity property system entirely (like we've discussed in-person), but I think that's a slightly larger project, so for now I'm using these methods.

src/pkg/caendr/caendr/models/datastore/hashable_entity.py
vlagrassa, line 1: Mostly pulled from obsolete `DataJobEntity` class

src/pkg/caendr/caendr/models/datastore/job_entity.py
```diff
@@ -94,29 +90,6 @@ def id(self):
 
 
 
-  ## Status ##
```
vlagrassa, line 97: Moved to `StatusEntity`

src/pkg/caendr/caendr/models/datastore/nemascan_mapping.py
```diff
@@ -1,47 +1,70 @@
...
-  ## Buckets & Paths ##
+  #
+  # Paths
+  #
 
-  # TODO: Overrides function in DataJobEntity parent class which isn't needed in this class.
-  #       Can this be merged with get_result_path for better inheritance?
-  def get_result_blob_path(self):
-    raise TypeError(f'Should not call get_result_blob_path on {self.__class__.__name__}')
+  # TODO: Move data files to Data bucket
```
vlagrassa, line 32: We should discuss standardizing the locations of certain data files for each tool. It's come up a bit in the past, but I have a much better idea now of what files are required by each tool and how we can divide them up / access them in a regular way

src/pkg/caendr/caendr/models/datastore/report_entity.py
```diff
@@ -0,0 +1,138 @@
...
+
+  @classmethod
+  def get_kind(cls):
+    return cls.kind
+
+
+  #
+  # Instantiation
+  # Route the GCPReport's create & lookup methods to the appropriate Entity methods
+  #
```
vlagrassa, line 43: I think the `Entity` class should also follow this factory method style of instantiation (`lookup` and `create`) -- part of a larger `Entity` redesign project for later

src/pkg/caendr/caendr/models/datastore/report_entity.py
```diff
@@ -0,0 +1,138 @@
...
+  # Data ID
+  # Subclass can designate a property field to use as the data ID
+  #
+
+  # The report field to use as the data ID
+  # Subclass should overwrite this abstract property,
+  # then the chosen prop value will be accessible through the getter below
+  @property
+  @abstractmethod
+  def _data_id_field(self) -> str:
```
vlagrassa, line 64: The data ID is the field I use to group Cloud Run jobs -- different submissions with the same data ID become different executions of the same job.

For the three tools, this is the data hash.  For database operations, this is the operation type (`DROP_AND_POPULATE_STRAINS`, etc.)

See `Report` class for more.

src/pkg/caendr/caendr/models/datastore/status_entity.py
vlagrassa, line 1: Mostly pulled from `JobEntity`

src/pkg/caendr/caendr/models/error.py
```diff
@@ -114,12 +114,10 @@ class UnprocessableEntity(InternalError):
 
 
 
-class CachedDataError(InternalError):
-  description = "This data has already been submitted by another user"
-  def __init__(self, report):
-    self.report = report
-    if report is not None and hasattr(report, 'data_hash'):
-      self.description = f'This data (hash {getattr(report, "data_hash")}) has already been submitted by another user'
+class JobAlreadyScheduledError(InternalError):
```
vlagrassa, line 117: I'm renaming this error type as part of the explicit separation between job creation & job submission.  Loosely speaking:
- If the user is trying to start a new job with data they've already submitted, that's a creation "error"
- If the user is trying to submit data that is already being computed (or already has results), that's a submission "error"

src/pkg/caendr/caendr/models/job_pipeline/database_operation_pipeline.py
```diff
@@ -98,7 +112,7 @@ def parse(cls, data, valid_file_extensions=None):
     # Package the data into report properties
     return {
       'props': {
-        'db_operation':   db_operation.value,
+        'db_operation':   db_operation,
```
vlagrassa, line 115: `Entity` and subclasses now handle `Enum`s, so we can keep the enum as-is

src/pkg/caendr/caendr/models/job_pipeline/heritability_pipeline.py
```diff
@@ -80,31 +93,107 @@ def parse(cls, data, valid_file_extensions=None):
...
       'props': data,
       'hash':  data_hash,
-      'files': [local_path],
+      'files': [local_file],
     }
 
 
+
+  #
+  # Fetching & Parsing Input
```
vlagrassa, line 111: Moved from `services`

src/pkg/caendr/caendr/models/job_pipeline/heritability_pipeline.py
```diff
@@ -80,31 +93,107 @@ def parse(cls, data, valid_file_extensions=None):
...
+    # If result exists and report status hasn't been marked as finished yet, update it
+    if result is not None and not self.is_finished():
+      self.report.set_status( JobStatus.COMPLETE )
+      self.report.save()
+
+    # Return the result
+    return result
+
+
+  def _parse_output(self, blob):
```
vlagrassa, line 149: Moved from `services`

src/pkg/caendr/caendr/models/job_pipeline/heritability_pipeline.py
```diff
@@ -80,31 +93,107 @@ def parse(cls, data, valid_file_extensions=None):
...
-  def upload(self, data_files):
 
-    # Heritability only takes one upload file
-    if len(data_files) > 1:
-      raise ValueError('Only one data file should be uploaded.')
+  def fetch_output(self, raw: bool = False):
+    result = super().fetch_output(raw=raw)
+
+    # If result exists and report status hasn't been marked as finished yet, update it
+    if result is not None and not self.is_finished():
```
vlagrassa, line 141: Moved out of the `view` function, so that from the caller's perspective this is handled implicitly

src/pkg/caendr/caendr/models/job_pipeline/indel_primer_pipeline.py
```diff
@@ -52,20 +65,141 @@ def parse(cls, data, valid_file_extensions=None):
     }
 
 
+
+  #
+  # Fetching & Parsing Input
+  #
+
+  def _parse_input(self, blob):
```
vlagrassa, line 73: Moved from `services`

src/pkg/caendr/caendr/models/job_pipeline/indel_primer_pipeline.py
```diff
@@ -52,20 +65,141 @@ def parse(cls, data, valid_file_extensions=None):
...
-  # File Storage
+  # Fetching & Parsing Output
   #
 
-  def upload(self, data_files):
 
-    # Indel Finder only takes one upload file
-    if len(data_files) > 1:
-      raise ValueError('Only one data file should be uploaded.')
+  def fetch_output(self, raw: bool = False):
```
vlagrassa, line 90: Moved from view function, so that from the caller's perspective this is handled implicitly

src/pkg/caendr/caendr/models/job_pipeline/indel_primer_pipeline.py
```diff
@@ -52,20 +65,141 @@ def parse(cls, data, valid_file_extensions=None):
...
+    # If result exists and report status hasn't been marked as finished yet, update it
+    if result is not None and not self.is_finished():
+      self.report.set_status( JobStatus.COMPLETE )
+      self.report.save()
+
+    # Return the result
+    return result
+
+
+  def _parse_output(self, blob):
```
vlagrassa, line 102: Moved from `services`

src/pkg/caendr/caendr/models/report/bucketed_report.py
vlagrassa, line 1: This class splits data files into three types, which can each have different buckets/paths:
- `report` - Any data specific to one report (e.g. input file)
- `data` - Any data used as input to multiple jobs (e.g. the lists of strains for each species)
- `work` - Any intermediate data generated by Nextflow

These in turn are defined using a `bucket` and a `prefix`, which are combined into the full `directory`.

src/pkg/caendr/caendr/models/report/gcp_report.py
vlagrassa, line 1: This subclass defines accessing report files on GCP (i.e. it's platform-specific)

src/pkg/caendr/caendr/models/report/report.py
vlagrassa, line 1: An abstract base class template for job report objects.  In this new organization, the `Report` object mediates all access to the datastore -- primarily locating/uploading/downloading files.

A lot of these methods are just templates for the methods defined in the `Entity` class.  By defining them here and implementing them there, we can provide a more general interface to the `JobPipeline` and its subclasses that isn't specific to GCP.

src/pkg/caendr/caendr/models/run/__init__.py
```diff
@@ -2,6 +2,3 @@
 from .base   import Runner
 from .gcp    import GCPRunner, GCPCloudRunRunner, GCPLifesciencesRunner
 from .local  import LocalRunner
-
-# Job-specific subclasses
```
vlagrassa, line 6: These are deprecated in favor of using the generic `Runner` class, and defining any parameters in the job-specific `JobPipeline` class.

The `JobPipeline` class primarily mediates between a given `Report` and a given `Runner`, e.g. reading props from the `Report` to construct the run environment.

src/pkg/caendr/caendr/models/run/gcp.py
```diff
@@ -135,54 +139,41 @@ def split_job_name(cls, job_name) -> (str, str):
...
-  def get_gcp_vars(self):
+  @classmethod
+  def default_environment(cls):
     return {
       "GOOGLE_SERVICE_ACCOUNT_EMAIL": sa_email,
       "GOOGLE_PROJECT":               GOOGLE_CLOUD_PROJECT_ID,
       "GOOGLE_ZONE":                  GOOGLE_CLOUD_ZONE,
     }
-  
-  def get_data_job_vars(self, report: DataJobEntity):
```
vlagrassa, line 159: This is moved to `JobPipeline` and `Report`

src/pkg/caendr/caendr/models/run/gcp.py
```diff
@@ -346,7 +372,7 @@ def _log_giveup(details):
   @backoff.on_exception(
       backoff.constant, SSLEOFError, max_tries=_MAX_TRIES_TOTAL, interval=20, jitter=None, on_backoff=_log_ssl_backoff.__func__, on_giveup=_log_ssl_giveup.__func__,
   )
-  def run(self, report, run_if_exists=False):
+  def run(self, command: list, env: dict, container_uri: str, params: dict, run_if_exists: bool = False) -> str:
```
vlagrassa, line 375: The `Runner` now takes these as arguments, since they're constructed by `JobPipeline` using data from `Report`

src/pkg/caendr/caendr/models/run/tools_gcp_cloudrun.py
vlagrassa, line 1: Deprecated in favor of using the generic parent class & filling in parameters using `JobPipeline` (and `Report`)

src/pkg/caendr/caendr/services/heritability_report.py
```diff
@@ -46,40 +40,3 @@ def get_heritability_reports(username=None, filter_errs=False):
   # Get list of reports and filter by date
   results = HeritabilityReport.query_ds(safe=not filter_errs, ignore_errs=filter_errs, filters=filters)
   return HeritabilityReport.sort_by_created_date(results, reverse=True)
-
-
-
-def fetch_heritability_report(report):
```
vlagrassa, line 52: Moved to `JobPipeline` / `Report`

src/pkg/caendr/caendr/services/indel_primer.py
```diff
@@ -134,14 +127,6 @@ def overlaps(s1, e1, s2, e2):
   return s1 <= s2 <= e1 or s2 <= s1 <= e2
 
 
-def fetch_ip_data(ip: IndelPrimerReport):
```
vlagrassa, line 137: Moved to `JobPipeline` / `Report`

src/pkg/caendr/caendr/services/indel_primer.py
```diff
@@ -172,94 +157,3 @@ def query_indels_and_mark_overlaps(species, strain_1, strain_2, chromosome, star
     results = [x for x in results if x['overlap'] is False]
     return sorted(results, key=lambda x: (x["START"], x["END"]))
   return []
-
-
-
-def fetch_indel_primer_report(report):
```
vlagrassa, line 178: Moved to `JobPipeline` / `Report`

src/pkg/caendr/caendr/services/validate.py
```diff
@@ -25,13 +25,13 @@ def get_delimiter_from_filepath(filepath=None, valid_file_extensions=None):
 
 
 
-def validate_file(local_path, columns, delimiter='\t', unique_rows=False):
+def validate_file(local_path_or_file, columns, delimiter='\t', unique_rows=False):
```
vlagrassa: Generalized to work on `os.PathLike`, so we can pass a `LocalFile` object directly

src/pkg/caendr/caendr/utils/file.py
```diff
@@ -12,17 +12,19 @@ def get_dir_list_sorted(path):
   return sorted([x for x in os.listdir(path) if not x.startswith(".")], reverse=True)
 
 
-def get_file_hash(filename, length=10):
-  ''' Generates the sha1 hash of a file's contents and returns the first 'length' characters '''
-  logger.debug(filename)
+def get_file_hash(path_or_file: os.PathLike, length=10):
```
vlagrassa, line 15: Generalized so we can pass a `LocalFile` object directly

src/pkg/caendr/caendr/utils/local_file.py
```diff
@@ -53,6 +56,18 @@ def __init__(self, local_file, valid_file_extensions=None):
...
 
+  #
+  # Local Path
+  #
+
+  @property
+  def local_path(self):
+    return self.__local_path
+  
+  def __fspath__(self):
```
vlagrassa, line 68: Required by `os.PathLike`

src/pkg/caendr/caendr/utils/local_file.py
```diff
@@ -69,7 +84,7 @@ def create(self):
       self.file.save(self.__local_path)
     except Exception as ex:
       raise self._make_save_error() from ex
-    return self.__local_path
+    return self
```
vlagrassa, line 88: Return object itself when entering context.  When combined with changes that generalize filepath handling to `os.PathLike`, this means we can pass the `LocalFile` object around and use it directly

src/pkg/caendr/caendr/utils/tokens.py
```diff
@@ -75,6 +76,25 @@ def __add__(self, other):
       raise ValueError(f'Cannot replace tokens in non-string value {other}')
 
 
+  def apply(f, *args, **kwargs):
```
vlagrassa, line 79: As mentioned above, the `TokenizedString` class ended up a bit complicated without great justification, so I'd like to streamline it when we have more time.  In the meantime, this method handles mapping a string-based function to a set of `TokenizedString` arguments, dealing w the tokens implicitly.

src/modules/site-v2/base/utils/tools.py
r-vieira: These explicit errors are incredibly helpful for consumers of this method. Consider adding python docs that describe the exceptions, e.g.:

```python
def my_function(arg1, arg2):

    """My function does something.

    Args:
        arg1 (int): The first argument.
        arg2 (str): The second argument.

    Raises:
        ValueError: If arg1 is not an integer.
        TypeError: If arg2 is not a string.
    """

    if not isinstance(arg1, int):
        raise ValueError("arg1 must be an integer.")
    if not isinstance(arg2, str):
        raise TypeError("arg2 must be a string.")

```

r-vieira: If the python docs are formatted correctly, then the IDE will show them when someone writes code that calls that method.

src/modules/site-v2/base/views/data/releases.py
```diff
@@ -139,7 +139,7 @@ def data_v02(params, files):
   browser_tracks_path = get_browser_tracks_path().get_string_safe()
   return {
     'browser_tracks_path': browser_tracks_path,
-    'browser_tracks_url': generate_blob_url(params['release_bucket'], browser_tracks_path),
+    'browser_tracks_url': generate_blob_uri(params['release_bucket'], browser_tracks_path, schema=BlobURISchema.HTTPS),
```
r-vieira, line 142: schema is 🔥

src/pkg/caendr/caendr/models/datastore/dataset_release.py
r-vieira: I like it. Btw -- [devil's advocate take] if the tokens are not needed after the production of the string, do we need to pass around TokenizedStrings, or would it be likely that we only need to assemble regular strings from the TokenizedString and then return regular strings instead?

src/pkg/caendr/caendr/models/datastore/hashable_entity.py
```diff
@@ -0,0 +1,46 @@
...
+class HashableEntity(Entity):
+  '''
+    An entity with a data hash property.
+  '''
+
+  ## Initialization ##
+
+  def __new__(cls, *args, **kwargs):
+    if cls is HashableEntity:
+      raise TypeError(f"Class '{cls.__name__}' should never be instantiated directly -- subclasses should be used instead.")
```
r-vieira, line 15: Although python doesn't support abstract classes natively, there is support via the `ABC` module, see:
https://docs.python.org/3/library/abc.html

Sample:
```python
from abc import ABC, abstractmethod

class AbstractClass(ABC):
    @abstractmethod
    def foo(self):
        print("foo")

try: 
    a = AbstractClass()
    a.foo()
except Exception as e:
    print(f"{e}") 
    # Can't instantiate abstract class AbstractClass with abstract methods foo
```

src/pkg/caendr/caendr/models/datastore/heritability_report.py
```diff
@@ -1,34 +1,51 @@
...
+      'TRAIT_FILE': self.input_filepath(schema=schema),
+    }
+
+
+  #
+  # Input & Output
+  #
+
+  _num_input_files = 1
+  _input_filename  = 'data.tsv'
```
r-vieira, line 42: most likely we will want to promote the `data.tsv` and `heritability_result.tsv` values to ENV vars.

src/pkg/caendr/caendr/models/datastore/nemascan_mapping.py
r-vieira: yes -- let's discuss it, this is a great suggestion!

src/pkg/caendr/caendr/models/datastore/nemascan_mapping.py
```diff
@@ -1,47 +1,70 @@
...
-  def get_report_blob_prefix(self):
-    return f'{self.get_result_path()}/{self.__report_path}'
+  #
+  # Input & Output
+  #
 
-  def get_input_data_path(self):
-    return f'{self.__input_data_path}'
+  _num_input_files = 1
+  _input_filename  = 'data.tsv'
```
r-vieira, line 58: same note as earlier, we should move this to a ENV var.

src/modules/site-v2/base/views/tools/pairwise_indel_finder.py
```diff
@@ -304,29 +303,26 @@ def report(id, file_ext=None):
...
-    result, format_table = modify_indel_primer_result(result)
-
-    # Update indel primer entity
-    # TODO: Is this the right time/place for this?
+    # Update indel primer empty status
     if ready:
-      if job.report['status'] != JobStatus.ERROR:
-        job.report['status'] = JobStatus.COMPLETE
-      job.report.empty = result is None  # TODO: Is this correct? I think empty should actually check if any rows exist in the df
+      job.report.empty = result['empty']  # TODO: 'empty' is no longer tracked as a prop. Should it be?
```
r-vieira, line 311: yes, a property would be preferable

r-vieira: in short, when the type-system can provide help, we should take it. When looking up keys in dicts, there is always a small change that a dict object might be missing a critical key. In such cases, creating class instances with all the defined properties can help mitigate the issue earlier. Some of this may come down to style preference, but an IDE might more likely detect an invalid property than it would detect an invalid (typo/etc) key.

src/pkg/caendr/caendr/models/datastore/report_entity.py
```diff
@@ -0,0 +1,138 @@
+# Built-ins
+from abc import abstractmethod
```
r-vieira, line 2: ah, had a comment about ABC earlier :)

r-vieira (approved):
wow! The new Job and Error handling classes bring a new level of sophistication to the platform. Well done!

PR #412, opened 2023-11-21
Hotfix: JS import for flashing message manually

vlagrassa:
Checks for `species_list` variable before trying to access.

I don't think this is broken in later branches (somehow), but it appeared for me on the Dataset Release page(s) in the current dev branch.


r-vieira (approved):
approved

PR #413, opened 2023-11-22
Enhancement/phenotype db database op

orzut:
Parsing functions for both bulk and non-bulk trait files.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -4,6 +4,7 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
+        
         {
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
vlagrassa, line 1: It's no problem here, but just a note for the future -- make sure to pay attention to what you're committing in the `launch.json` file.  When I change e.g. the database access method for testing, I usually don't commit those changes, since it doesn't necessarily make sense for everyone's `launch.json` to point to my local values.

In this case, I think it makes sense to connect to a test file by default, as an added layer of "security" -- i.e. if someone runs the project in the future without reading carefully, they won't automatically connect to the central database and potentially wipe out data.

vlagrassa (approved):
Looks good! Just left one comment about the `launch.json` file -- it doesn't break anything, so I'll go ahead with the merge.

PR #414, opened 2023-11-27
Main environment public bucket

vlagrassa:
Point QA and dev environments to the main public bucket.

Also adds some support for fallback environment variables.  This ended up being unnecessary for this PR, but I've left it in just in case.

Marking as draft until we figure out IAM permissions for the service account(s).


src/pkg/caendr/caendr/utils/env.py
```diff
@@ -72,6 +72,44 @@ def get_env_var(key, value=None, can_be_none=False, as_template=False, var_type=
   return v
 
 
+
+def get_env_var_with_fallback(key, *fallback_keys, fallback_on_err=True, **kwargs):
```
r-vieira, line 76: I appreciate the dynamic nature of this lookup, but worry that it introduces a level of implicit behaviors that might hide issues across different parts of the code base. Instead, if we need to revise the ENV vars, I would prefer we make it clearer at that level and set the protocol for *which* bucket should be used for various ends. Let's discuss.

r-vieira (approved):
approved.

src/pkg/caendr/caendr/utils/env.py
vlagrassa: Hm, fair point. As far as the buckets go, I wonder if it would be better to replace generic bucket vars like `PUBLIC_BUCKET` and `PRIVATE_BUCKET` with specific ones like `DATASET_RELEASE_BUCKET` or `PHOTOS_BUCKET`.  We could maybe talk about setting defaults (e.g. `DATASET_RELEASE_BUCKET` falls back to `PUBLIC_BUCKET`), but to your point it would be better to define all the bucket variables explicitly.

r-vieira: I really like this suggestion! Let's add this to the next cycle.

r-vieira (approved):
approved

PR #415, opened 2023-11-28
Create Datastore Entity kind for Trait Files

vlagrassa:
- Creates new `TraitFile` subclass of `Entity`
- Creates new `PublishStatus` enum (draft, should discuss the specifics)
- Pulls out / creates a few more `Entity` template subclasses:
  - `FileRecordEntity` - an `Entity` that tracks a file + metadata (`TraitFile`, `GeneBrowserTracks`)
  - `SpeciesEntity` - an `Entity` with a species field (for now, `TraitFile` and `DatasetRelease`)
  - `PublishableEntity` - an `Entity` with a `PublishStatus` field
- Cleans up `Entity`:
  - Explicit serialization of fields
  - Organizes `__init__.py` based on dependencies
- Updates a few blob get/check functions to be more in line with new style


r-vieira (approved):
approved

PR #417, opened 2023-11-30
adding missing classes to __init_.py

orzut:


src/pkg/caendr/caendr/models/datastore/__init__.py
```diff
@@ -16,6 +16,7 @@
 from .user_owned_entity   import UserOwnedEntity     # Imports User
 from .hashable_entity     import HashableEntity
 from .report_entity       import ReportEntity        # Subclasses JobEntity, UserOwnedEntity, as well as GCPReport
+from .deletable_entity    import DeletableEntity     # Subclasses Cart
```
vlagrassa: I think the subclass relationship actually goes the other way: `DeletableEntity` stands on its own, and `Cart` subclasses from `DeletableEntity`.

Then when you read the comments in the `__init__` file, you want to make sure everything listed in `# Subclasses XYZ` is defined higher up in the file.

vlagrassa (changes requested):

src/pkg/caendr/caendr/models/datastore/__init__.py
vlagrassa: Sorry, just realized my inline comments might be a bit grammatically unclear.  I'm using "subclasses" informally as a verb to mean "is a subclass of", not a noun meaning "here's a list of subclasses".  Now that I think about it though, that might not be super clear.

This format mimics the structure of the Python class list in the individual files (i.e., in `cart.py`, you'd see that the `Cart` class is a subclass of `DeletableEntity`).

orzut: Fixed in the commit 3d656142e55fa16429427906cbcdc541a1749b83

vlagrassa (approved):

r-vieira (approved):
approved

PR #416, opened 2023-11-29
Rework ETL Pipeline (prep work for SQL PhenotypeDB)

vlagrassa:
Splitting the larger branch into two sections:
1) Reworking the SQL ETL pipeline, with an eye toward adding a PhenotypeDB table (this PR)
2) Adding a config object & parsing functions for the PhenotypeDB table (incoming)

This PR is largely me paying off my technical debt, now that I have a better understanding of all the systems at play.  Large parts of the ETL pipeline were overly complicated and/or based on project parameters that have changed (e.g. downloading external files from WormBase directly).

Highlights:
- Updating another blob function (`download_blob_to_file` this time):
  - Variadic `*path`, like the get/check functions
  - Uses Werkzeug's `secure_filename` on the destination filename
  - Updates all places where this function is used (incl. the Indel Primer module -- will want to test thoroughly)
- Uses `os.path.join` where appropriate, instead of direct concatenation w `'/'`
- Starts to build out different types of `local_file` objects:
  - Existing `LocalFile` rebranded as `LocalUploadFile`, to manage a file upload from Flask
  - New `LocalDatastoreFile` class manages downloading/caching a file from GCP (for db table source files)
  - Start on `ForeignResourceWatcher` abstract base class (originally `SpeciesDict`) to manage the same datastore file "template" across the different species
- Handle ETL operations (drop, create, populate) in `ETLManager`
  - Removes old `DatasetManager` class, since tracking/downloading non-local resources is now handled by `LocalDatastoreFile` objects
- New `TableConfig` class with global objects representing each table
  - Track the list of resources to download + parsing function
  - Used directly in db operations preflight check to make sure all resources exist
  - Unifies `Strain` table with the other tables

PR #418, opened 2023-12-05
SQL PhenotypeDB

vlagrassa:
Note: Marking as draft until #416 is merged. Once merged, the changes in this branch should only be the ones outlined below.

--

Joint effort with @orzut :)

Highlights:
- Add support for `PhenotypeDB` table:
  - Multiple parsing functions (bulk files like Zhang vs. single files)
  - `TableConfig` object
  - Updated type annotations
  - `DbOp` support (can run as a job)
- Add new `ParseConfig` class used by existing `TableConfig` -- allows one table to use different parsing functions on different files when building
- Rework `ForeignResource` template class
  - Separate `ForeignResource` and `ForeignResourceTemplate`
  - Bring fetching / building / etc under one interface (for datastore files & Google Sheets)
- Add explicit `RELOAD_FILES` env var


.vscode/launch.json
```diff
@@ -54,12 +55,12 @@
                 "ENV": "main",
                 "FLASK_APP": "main.py",
                 "FLASK_ENV": "development",
-                "MODULE_DB_OPERATIONS_CONNECTION_TYPE": "localhost",
-                // "MODULE_DB_OPERATIONS_CONNECTION_TYPE": "file",
-                // "MODULE_DB_OPERATIONS_CONNECTION_FILE": "${workspaceFolder}/temp/local.db",
+                // "MODULE_DB_OPERATIONS_CONNECTION_TYPE": "localhost",
```
r-vieira, line 58: as part of ongoing/persistent documentation, we should keep a comment in this file that tracks all the possible values for the connection type: // Supported values: 'localhost', 'file', ...

r-vieira (approved):
looks good!

.vscode/launch.json
vlagrassa: Added in bbb849ad1453a92d84a27952f66e11f831ca1014

PR #420, opened 2023-12-05
Better error checking if Dataset Release files don't exist

vlagrassa:
Specifically, if the `DatasetRelease` entity exists but *none* of the files exist, would crash trying to render Markdown files.

The existing `files and ...` guards work when `files is None` (render fallback) or `files` is an object with fields (i.e. some files exist, but maybe not the desired MD file).

The problem was, if `files` is an empty object `{}`, then `{} and {}.get(...)` will evaluate to `{}` (it takes precedence over the `None` result in the second argument slot).  This would pass an empty object to `render_ext_markdown` as the `url`, which in turn wasn't properly caught (since that function would only guard against `None`).

Changes:
- Better guard for files: `files or {}` will fallback to an empty dict, then `.get` will retrieve `None` since no fields exist.
- Better guard for empty "url"s: now checks against `bool` conversion, which catches `None` as well as `''` and `{}`.

### DIFF
--- a/src/modules/site-v2/templates/data/v2.html
+++ b/src/modules/site-v2/templates/data/v2.html
@@ -48,7 +48,7 @@
                     <div class="row">
                         <div class="col-6">
                             <h2>Release Notes</h2>
-                            <p>{{ render_ext_markdown(files and files.get('release_notes'), ignore_err=true, backup_text="*Release notes are not available at this time.*") }}</p>
+                            <p>{{ render_ext_markdown((files or {}).get('release_notes'), ignore_err=true, backup_text="*Release notes are not available at this time.*") }}</p>
                         </div>
                         <div class="col-6">
                             <div class="card">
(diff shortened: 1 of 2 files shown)


src/modules/site-v2/templates/data/v2.html
```diff
@@ -48,7 +48,7 @@
                     <div class="row">
                         <div class="col-6">
                             <h2>Release Notes</h2>
-                            <p>{{ render_ext_markdown(files and files.get('release_notes'), ignore_err=true, backup_text="*Release notes are not available at this time.*") }}</p>
+                            <p>{{ render_ext_markdown((files or {}).get('release_notes'), ignore_err=true, backup_text="*Release notes are not available at this time.*") }}</p>
```
r-vieira, line 51: do we need an error message in case files is empty?

r-vieira (approved):
approved

PR #421, opened 2023-12-06
Draft Phenotype DB Pipeline

vlagrassa:
We can now run[1] a phenotype database/comparison/analysis job!  For now, I compute the correlation value(s) as part of `PhenotypePipeline.parse_output`, which is analogous to how other jobs compute simple inline values from their output files.  I'm also drawing the data from the datastore files, but we'll want to work on reading from SQL instead.

[1] sort of!

I set up a few test pages under a `phenotype-comparison` blueprint.  None of these are front-facing, or even drafts of front-facing pages -- they're just for me to make sure the pipeline works.  This PR does not register that blueprint; instead, I'll work on merging that code into the page templates provided by @natalieroman.

Highlights:
- Client-side scatterplot histogram code (D3.js)
- Draft phenotype pipeline classes:
  - `PhenotypeReport` datastore entity
  - `PhenotypePipeline` to run job
- Draft running analysis:
  - Spearman function runs "inline"
  - Running the report is folded into `parse_input` and `parse_output`, by analogy with other jobs
- Misc. utility updates


src/modules/site-v2/base/static/js/scatter-histogram.js
```diff
@@ -0,0 +1,253 @@
...
+/* Add a histogram along an axis of a graph.
+ *
+ * Arguments:
+ *   - d (int): The dimension of the histogram. 0 for x-axis, 1 for y-axis
+ *   - axis: A d3 scaleLinear object mapping the domain of the given axis to its position in the graph
+ *   - data: The data to plot
+ *   - target: The SVG element to create the histogram in
+ *   - config: Optional keyword arguments.
+ */
+function add_histogram_along_axis(d, axis, data, target, config) {
```
r-vieira, line 38: optional for the `config` object, consider a default value, `config={}` with the correct data shape.

src/modules/site-v2/base/static/js/scatter-histogram.js
```diff
@@ -0,0 +1,253 @@
...
+
+  // Validate the dimension parameter
+  if (!(d === 0 || d === 1)) {
```
r-vieira, line 41: Optional, since this is part of the API for the histogram, but consider upgrading `0` and `1` to enums to avoid 'magic numbers'. 
```
const HistogramDimension = {
  'x_axis': 0,
  'y_axis': 1
};
```

src/modules/site-v2/base/static/js/scatter-histogram.js
```diff
@@ -0,0 +1,253 @@
...
+  // and translate along the primary axis so they all sit next to one another
+  const bar = graph.selectAll('.bar')
+    .data(bins)
+    .enter().append('g')
+    .attr('class', 'bar')
+    .attr('transform', function(n) {
+
+      // D3 draws from top left, so along the x-axis,
+      // offset the y-position of the bar based on the white space that should be above it
+      if (d == 0) return `translate(${ axis(n.x0) }, ${ bin_val_to_height(max_bin_amount - n.length) })`;
```
r-vieira, line 91: consider using triple equals `===` on these comparisons to avoid type coercion side effects (eg: `false == 0` is `true`)

src/modules/site-v2/base/static/js/scatter-histogram.js
```diff
@@ -0,0 +1,253 @@
...
+      .style('stroke', stroke_color);
+
+    // Create tooltip for data point mouseover
+    const tooltip = d3.select("#phenotype-chart")
+      .append("div")
+      .attr("class", "tooltip")
+      .style("opacity", 0);
+
+    // Add mouseover listener for individual dots
+    dots.on('mouseover', function(d) {
```
r-vieira, line 216: hum....on mobile we have `touchstart|touchmove|touchend` -- is there parity of experience we can have on mobile/touch devices with the charts?

src/modules/site-v2/base/views/tools/phenotype_comparison.py
```diff
@@ -0,0 +1,138 @@
...
+@phenotype_comparison_bp.route('/submit', methods=["POST"])
+@jwt_required()
+def submit():
+  form = PhenotypeComparisonForm(request.form)
+  user = get_current_user()
+
+  # Validate form fields
+  # Checks that species is in species list & label is not empty
+  if not form.validate_on_submit():
+    msg = "Invalid submission"
```
r-vieira, line 65: if we can be more descriptive to the end-user on specifics on why the submission failed, we should consider it.

src/modules/site-v2/templates/tools/phenotype/report.html
```diff
@@ -0,0 +1,58 @@
...
+{% block script %}
+<script src="{{ url_for('static', filename='js/scatter-histogram.js') }}"></script>
+
+<script>
+
+const data = {{ data.trait_values|tojson }};
+
+scatterplot_histograms('#phenotype-chart', data, {
+    hist_height:   60,
+    width:         600,
```
r-vieira, line 43: second order priority -- we might need to adjust width/height for mobile devices. (The first priority is to get the laptop version).

src/modules/site-v2/templates/tools/phenotype/submit.html
```diff
@@ -0,0 +1,71 @@
...
+  </div>
+
+{% endblock %}
+
+
+{% block script %}
+{% from "_includes/strain-listing.html" import unpack_species_list  %}
+{% from "_scripts/submit-job.js"        import def_submit_job       %}
+
+<script>
```
r-vieira, line 28: prefer `<script type="module">` to keep scopes local to the script tag.

src/modules/site-v2/templates/tools/phenotype/submit.html
```diff
@@ -0,0 +1,71 @@
...
+  {% include '_scripts/utils.js' %} {#/* defines: flash_message */#}
+  {{ def_submit_job('phenotype_comparison', true) }}
+
+
+  /* Species */
+  const species_list = {{ unpack_species_list(species_list, species_fields) }};
+  let species;
```
r-vieira, line 35: not used?

src/modules/site-v2/templates/tools/phenotype/submit.html
```diff
@@ -0,0 +1,71 @@
...
+
+  /* Traits */
+  let traits;
+
+
+  // Retrieve trait lists
+  fetch_json("{{ url_for('api_trait.query_all', split_by_species=True) }}")
```
r-vieira, line 42: are we using the `traits` elsewhere? 

If we are just fetching_json() initially without argument mutations, we might be able to include the data during template rendering phase.

r-vieira (approved):
Looks great! See inline comments.

src/modules/site-v2/templates/tools/phenotype/submit.html
vlagrassa: We aren't yet, but we might need it, depending on the implementation of the HTML table.  If this request ends up being unnecessary, I'll remove it.

src/modules/site-v2/templates/tools/phenotype/submit.html
vlagrassa: Not yet -- we'll want to do some species verification based on the selected trait(s).  I'll have a better idea how to do that once we figure out the trait table.

(Noted in Basecamp)

src/modules/site-v2/templates/tools/phenotype/report.html
vlagrassa: Agreed -- I imagine we'll want this to choose this value based on window width (or maybe even make the graph size responsive).  I've just filled in reasonable values for now, and still have to work w Natalie to lay out the graph(s) on the report page.

src/modules/site-v2/base/static/js/scatter-histogram.js
vlagrassa: Fixed in 7651b21e7c0ec4105403c84ddef0a8bb07ff348c

src/modules/site-v2/base/static/js/scatter-histogram.js
vlagrassa: Added in 7651b21e7c0ec4105403c84ddef0a8bb07ff348c

src/modules/site-v2/base/static/js/scatter-histogram.js
vlagrassa: Good question -- moving to Basecamp

src/modules/site-v2/base/static/js/scatter-histogram.js
vlagrassa: Added in 6a394b70889c8c650f042f4401607b486f82fbc6

src/modules/site-v2/templates/tools/phenotype/submit.html
vlagrassa: Fixed in 228841d37b4de9ea264ee8aa00fa3d82c2344aed

vlagrassa: Fixed in 228841d37b4de9ea264ee8aa00fa3d82c2344aed

src/modules/site-v2/base/views/tools/phenotype_comparison.py
vlagrassa: For our records -- this method of receiving / validating the submission has been replaced with a JSON solution.  Will still be worth testing invalid submission values.

See [current solution](https://github.com/AndersenLab/CAENDR/blob/191763d0cb7a9284d074735a545c3c02dd732c59/src/modules/site-v2/base/views/tools/phenotype_database.py#L121) (as of this comment).

PR #422, opened 2023-12-08
Added "Zheng et al." tab to tables

natalieroman:
Added tab to the phenotype table per discussion with @vlagrassa


r-vieira (approved):
looks great!

PR #425, opened 2023-12-13
Hotfix: BAM/BAI download page schema argument

vlagrassa:
When using `generate_blob_uri`, the `schema` must be passed as a keyword arg.

I checked all other uses of `generate_blob_uri` and they all pass the argument correctly.


r-vieira (approved):
good

PR #429, opened 2023-12-15
Finalize(?) TraitFile entity props set, seed entities from Google Sheets

vlagrassa:
Notes:
- This PR adds the secret `ANDERSEN_LAB_TRAIT_SHEET`
- Still waiting on trait files with the correct columns / filenames from Andersen Lab. Once we receive those, we can upload to GCP and run the new `POPULATE_PHENOTYPES_DATASTORE` database operation :)

Highlights:
- Updates `TraitFile` props set:
  - Split `trait_name` into `trait_name_caendr` and `trait_name_user`
  - Add `dataset` (for now, `caendr` or `zhang`) and use to construct filepath in GCP
  - Update `doi` to more general `publication`
  - Add `institution`, `capture_date`, and `tags`
- Update trait file location in GCP. Path is now `trait_files/{ dataset }/{ species }/`
- Add db op to load / update `caendr` trait file datastore entities from the Google Sheet
  - New `DbOp` value `POPULATE_PHENOTYPES_DATASTORE`
  - New secret value `ANDERSEN_LAB_TRAIT_SHEET`
- Move `DbOp` enum definition to its own file:
  - Now lives in `models/sql`, which better reflects scope
  - Now used in `db_operations` module to verify the `DATABASE_OPERATION` env var


src/modules/db_operations/seed_trait_files.py
```diff
@@ -0,0 +1,96 @@
...
+
+# Get secret value(s)
+ANDERSEN_LAB_TRAIT_SHEET = get_secret(f'ANDERSEN_LAB_TRAIT_SHEET')
+
+
+
+def populate_andersenlab_trait_files():
+  '''
+    Create / update datastore entities for each record in the trait list Google Sheet.
+    Assumes the following:
```
r-vieira, line 20: we should explicitly document that this function does not remove rows from datastore if a ROW is removed from the sheets after having been synced to datastore.

r-vieira (approved):
looks great!

PR #431, opened 2023-12-18
Creating PhenotypeMetdata SQL table

orzut:
- Created PhenotypeMetadata SQL table which gets seeded by TraitFile entities from GCP Datastore using a parsing function.
- Added relationship to PhenotypeDatabase, where they join on unique trait names
- Created queries for PhenotypeMetadata


src/modules/db_operations/operations.py
```diff
@@ -112,26 +115,43 @@ def drop_and_populate_phenotype_db(app, db, species, reload_files=True):
   logger.info("Loading phenotypes...")
   etl_manager.load_tables(PhenotypeDatabase, species_list=species)
 
+def drop_and_populate_phenotype_metadata(app, db, species, reload_files=True):
```
vlagrassa, line 118: Quick question, is it possible to drop & populate the two phenotype tables separately? If you're using foreign keys in one of them, it might prevent the other table from being cleared correctly.  If you've tested it and it works that's great :)

src/pkg/caendr/caendr/api/phenotype.py
vlagrassa, line 1: Love the API querying functions!

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
```diff
@@ -0,0 +1,70 @@
...
+    Parsing function to load PhenotypeMetadata table from datastore.
+    For bulk file ouputs each trait as a new row and replicates metadata for all of them
+  """
+  
+  logger.info('Parsing function for traits metadata')
+  
+  for fname, fpath in files.items():
+
+    # Get metadata for the file 
+    md = fpath._metadata
```
vlagrassa: Quick note on this -- you should be able to get the metadata using `fpath.metadata()`.

Generally in Python, any class method / variable that starts with `_` is considered "private" -- it's a way of signaling to other programmers that you should only access it inside the class definition itself.  Technically in Python, there's no such thing as "private" vs. "public" fields, since the language itself makes all fields public.  However, having "private" fields is often useful to separate implementation from access.

(For example, I might decide later that the internal `_metadata` field needs to change, which could break this code; but, as long as I keep the return value of the external `metadata()` function the same, I can change as much as I want without breaking your code.)

In this case, I've marked the internal `_metadata` field as private for reasons to do with the parent class of `LocalDatastoreFile` -- happy to explain more in-depth if you'd like.

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
```diff
@@ -0,0 +1,70 @@
...
+             'species':           md.species.name,
+             'wbgene_id':         row[1],
+             'description_short': md['description_short'],
+             'description_long':  md['description_long'],
+             'units':             md['units'],
+             'publication':       md['publication'],
+             'protocols':         md['protocols'],
+             'source_lab':        md['source_lab'],
+             'institution':       md['institution'],
+             'submitted_by':      md.serialize()['submitter'],
```
vlagrassa: Ah, for this field, you might want to check out `md['username']` or `md.get_user()`.  That way you can access the `User` object directly.

I add the `submitter` field when serializing because I lose access to certain functions when I convert the TraitFile object to a JSON object, but I want to use those functions to compute a value on the front-end.  Since this function is on the back-end, you have direct access to the TraitFile object and its functions, so you can use them directly.

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
```diff
@@ -0,0 +1,70 @@
...
+      'species':           md.species.name,
+      'wbgene_id':         'N/A',
+      'description_short': md['description_short'],
+      'description_long':  md['description_long'],
+      'units':             md['units'],
+      'publication':       md['publication'],
+      'protocols':         md['protocols'],
+      'source_lab':        md['source_lab'],
+      'institution':       md['institution'],
+      'submitted_by':      md.serialize()['submitter'],
```
vlagrassa: Same comment as above

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
```diff
@@ -0,0 +1,70 @@
...
+
+def parse_phenotype_metadata(species: Species, **files: LocalDatastoreFile):
+  """
+    Parsing function to load PhenotypeMetadata table from datastore.
+    For bulk file ouputs each trait as a new row and replicates metadata for all of them
+  """
+  
+  logger.info('Parsing function for traits metadata')
+  
+  for fname, fpath in files.items():
```
vlagrassa: Quick note on naming -- technically, `files.items()` will give you a list of file names (of type `str`) and file objects (of type `LocalDatastoreFile`), as opposed to just paths.  This isn't a huge deal, but it might be clearer to use e.g. `fname, f` or `file_name, file_object` or something similar.  Might make the usage of `fpath.metadata()` below a bit clearer.

vlagrassa (changes requested):
Looks good!  All the database stuff looks great.  Left a few quick notes mainly on accessing attributes -- let me know if I can clear anything up.

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
orzut: Thanks for educating on this! Fixed in the commit 6dff197494274095938bbf64cef3d2accce75ea2

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
orzut: Fixed in the commit https://github.com/AndersenLab/CAENDR/commit/6dff197494274095938bbf64cef3d2accce75ea2

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
orzut: Fixed in the commit https://github.com/AndersenLab/CAENDR/commit/6dff197494274095938bbf64cef3d2accce75ea2

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
orzut: Fixed in the commit https://github.com/AndersenLab/CAENDR/commit/6dff197494274095938bbf64cef3d2accce75ea2

src/modules/db_operations/operations.py
orzut: That's a good question. I would weigh on your and @r-vieira decision how we want to handle this. As of now, I think it fails if we won't populate tables in the right order (PhenotypeMetadata, PhenotypeDatabase).

vlagrassa: Ah, in that case, I'd recommend combining them into a single database operation -- see `drop_and_populate_wormbase_genes` for an example, and the doc comment in `ETLManager.clear_tables` for more info on handling multiple tables with dependencies.  That way, Andersen Lab would just run a single operation that handles the entire rebuild process for them.

vlagrassa: I'd defer to @r-vieira though, if there's ever a situation where we would want to drop / rebuild them separately?

vlagrassa (approved):
Looks great!

(Good catch that `metadata` is a `@property` instead of a regular method 👍)

src/pkg/caendr/caendr/api/phenotype.py
```diff
@@ -0,0 +1,48 @@
+import os
+
+from caendr.models.datastore import Species
+from caendr.models.error import BadRequestError
+from caendr.models.sql import PhenotypeMetadata
+from caendr.services.cloud.postgresql import rollback_on_error
+
+
+@rollback_on_error
+def query_phenotype_metadata(is_bulk_file=False, include_values=False, species: str = None):
```
r-vieira, line 10: thinking of the order and topology of the arguments -- the api appears to support a single species or `None` for all species -- should it also support multiple species (eg: elegans e tropicalis)?

src/pkg/caendr/caendr/api/phenotype.py
```diff
@@ -0,0 +1,48 @@
+import os
+
+from caendr.models.datastore import Species
+from caendr.models.error import BadRequestError
+from caendr.models.sql import PhenotypeMetadata
+from caendr.services.cloud.postgresql import rollback_on_error
+
+
+@rollback_on_error
```
r-vieira, line 9: consider removing the rollback -- `@rollbacks` are great when there are side-effects to the DB calls and we want to bundle them into a transaction in order to undo the changes in case something goes wrong. In this case, it appears that we are just `reading` (query.all) and wouldn't get into the scenario where we would need to rollback a change to the db.

src/pkg/caendr/caendr/services/sql/etl/phenotype_metadata.py
```diff
@@ -0,0 +1,70 @@
...
+  
+  for file_name, file_object in files.items():
+
+    # Get metadata for the file 
+    md = file_object.metadata
+
+    # If bulk file, open it and yield each trait as a new row with the file's metadata
+    if md.is_bulk_file:
+      with open(file_object) as csv_file:
+        for idx, row in enumerate( csv.reader(csv_file, delimiter='\t') ):
```
r-vieira, line 25: it might be useful to document here the expected order of items in the `row`. 

Also, consider unpacking the row into named local vars to ensure we have the correct expectations, eg: 
```
row = [1,2]
foo, bar = row 
```

then use the named variables to replace places where we have `row[1]`.

r-vieira (approved):
Looks great! see inline comments.

PR #432, opened 2024-01-02
Front-end Phenotype table

orzut:
**Populating front-end tables with SQL data.**
I'm using Datatables' `server-side processing` only for rendering bulk files (i.e. fetching data only by request).
As for now, for non-bulk files I'm rendering the data using Jinja variables. Once we have all non-bulk files in place I'll adapt the front-end rendering.

_Note:_ We haven't populated SQL tables yet, so the `phenotype-database` page will throw 500 error. (Would appreciate any ideas how to handle it until we have populated tables)


src/pkg/caendr/caendr/api/phenotype.py
```diff
@@ -46,3 +46,11 @@ def get_all_traits_metadata():
     """
     return PhenotypeMetadata.query.all()
 
+
+def get_trait(trait_name, include_values=False):
+   query = PhenotypeMetadata.query.filter_by(trait_name=trait_name)
```
vlagrassa, line 51: If this is the primary key, I think you can use `query.get` in place of `filter_by`.  If not, ignore this message :)

r-vieira:
> **Populating front-end tables with SQL data.** I'm using Datatables' `server-side processing` only for rendering bulk files (i.e. fetching data only by request). As for now, for non-bulk files I'm rendering the data using Jinja variables. Once we have all non-bulk files in place I'll adapt the front-end rendering.
> 
> _Note:_ We haven't populated SQL tables yet, so the `phenotype-database` page will throw 500 error. (Would appreciate any ideas how to handle it until we have populated tables)

Good question -- Once the SQL tables are created, our queries should return an empty result set (instead of 500), and the frontend should have a default message when there is no content to be rendered. Regarding the populating the tables -- for lower environments, we can load sample/mock data from a sample local json file until we can cascade production data from prod to the lower envs. 

Lmk if this helps.

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
 def phenotype_database():
+  """
+    Phenotype Database table (non-bulk)
+  """
+  # Get the list of traits for non-bulk files
+  try:
+    traits_non_bulk = query_phenotype_metadata()
+  except Exception as ex:
+    logger.error(f'Failed to retrieve the list of traits: {ex}')
+    abort(500)
```
r-vieira, line 74: consider passing a `description=` in the `abort(...)` so that we can use it in the front-end to hint to the end-user what happened.

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
+  """
+    Phenotype Database table (bulk)
+    Fetch table content by request for each page and render the data for Datatables()
+  """
+  try:
+    # get parameters for query
+    draw = request.args.get('draw', type=int)
+    start = request.args.get('start', type=int)
+    length = request.args.get('length', type=int)
+    search_value = request.args.get('search[value]', '').lower()
```
r-vieira, line 98: we should add a guard here to validate *all* the user input and assert that it is in valid ranges/values/lists/etc. If the request can't be processed, we should catch it before tying to process it and notify the user with a proper error (e.g.: 400 bad request)>

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
+
+    query = PhenotypeMetadata.query
+    if search_value:
+      query = query.filter(
```
r-vieira, line 102: while this is idiomatic in python, prefer not reusing the same variables to hold different types of data -- in this  case, the result from `query.filter()` might be closer to a `results` or `response` variable.

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -228,3 +309,15 @@ def report(id):
 
     'JobStatus': JobStatus,
   })
+
+def convert_joined_query_tojson(q):
+  """
+    Convert PhenotypeMetadata query joined with PhenotypeDatabase to JSON
+  """
+  json_data = []
+  for trait in q:
```
vlagrassa, line 318: Not mandatory, but --

This JSON conversion seems like a pretty basic & useful operation, so if you want, you can actually move it to the `PhenotypeMetadata` class itself.  Might look something like:

```
def to_json(self):
    json_trait = super().to_json()  # Change to super to avoid infinite recursion
    phenotype_values = [ v.to_json() for v in self.phenotype_values ]
    json_trait['phenotype_values'] = phenotype_values
    return json_trait
```

You could then directly call it with `trait.to_json()` for each `trait` in your list.

Up to you whether this makes sense.  And if it does, whether it should be the default `to_json` behavior, or whether it should be a new method (like `to_json_with_values` or something).

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
+@phenotype_database_bp.route('/traits-list')
+@cache.memoize(60*60)
+@compress.compressed()
+def get_traits_json():
+  """
+    Get traits data for non-bulk files in JSON format (include phenotype values)
+  """
+  try:
+    traits_query = query_phenotype_metadata()
+    traits_json = [ trait.to_json() for trait in traits_query]
```
vlagrassa, line 143: I think this might be a duplicate line from before `convert_joined_query_tojson`?

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
+    data = query.offset(start).limit(length).from_self().\
+      join(PhenotypeMetadata.phenotype_values).all()
+    
+    json_data = convert_joined_query_tojson(data)
+    
+    total_records = query.count()
+
+    response_data = {
+        "draw": draw,
+        "recordsTotal": total_records,
```
r-vieira, line 124: if possible, we should use snake-case for json payloads. (eg: `"total_records": total_records, `)

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
+        "recordsFiltered": total_records,
```
r-vieira, line 125: is there a semantic difference between `total_records` and `filtered_records`? If so, should we get the `total_records` before the `.filter()`?

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
+        "data": json_data
+    }
+
+  except Exception as ex:
+    logger.error(f'Failed to retrieve the list of traits: {ex}')
+    response_data = []
```
r-vieira, line 131: is it clear to the end-user that this was an error (anything that throws inside the try/catch block) vs an empty result set? If appropriate, consider propagating the error message (user friendly version) to the end-user.

vlagrassa (approved):
This looks really good! :)

Posted one thought on the JSON conversion -- if it makes sense, you could move the code to a method in the `PhenotypeMetadata` class itself, then `to_json` on each query result would work more like you'd want it to.  (It wasn't intuitive to me at first that you could write methods like that, so I figure I should mention it here in case you find it useful).

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -59,12 +63,89 @@ def results_columns():
...
+def get_traits_json():
+  """
+    Get traits data for non-bulk files in JSON format (include phenotype values)
+  """
+  try:
+    traits_query = query_phenotype_metadata()
+    traits_json = [ trait.to_json() for trait in traits_query]
+    traits_json = convert_joined_query_tojson(traits_query)
+  except Exception:
+    traits_json = []
```
r-vieira, line 146: same idea as before -- should we differentiate errors vs empty result sets?

src/modules/site-v2/base/views/tools/phenotype_database.py
vlagrassa: (Just realized -- you don't have to move the `super()` call to the start of the method.)

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -228,3 +309,15 @@ def report(id):
 
     'JobStatus': JobStatus,
   })
+
+def convert_joined_query_tojson(q):
```
r-vieira, line 313: add pythondocs to describe the input/outputs of this function

r-vieira (approved):
looks good! see inline comments

PR #437, opened 2024-01-10
Front-end Phenotype table

orzut:
- Fixes for the previous PR
- Displaying Categories tags programmatically
- Adding Filtering by categories for non-bulk traits
- Managing fields in the side bar (hide if there's no value)
- [Request Strains minor change] Noticed Delete button alignment mismatches the rest of the table

### DIFF
--- a/src/modules/site-v2/base/views/tools/phenotype_database.py
+++ b/src/modules/site-v2/base/views/tools/phenotype_database.py
@@ -115,7 +115,7 @@ def get_zhang_traits_json():
     data = query.offset(start).limit(length).from_self().\
       join(PhenotypeMetadata.phenotype_values).all()
     
-    json_data = convert_joined_query_tojson(data)
+    json_data = [ trait.to_json_with_values() for trait in data ]
     
     total_records = query.count()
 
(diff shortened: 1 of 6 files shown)


src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -140,8 +142,7 @@ def get_traits_json():
   """
   try:
     traits_query = query_phenotype_metadata()
-    traits_json = [ trait.to_json() for trait in traits_query]
-    traits_json = convert_joined_query_tojson(traits_query)
+    traits_json = [ trait.to_json_with_values() for trait in traits_query]
   except Exception:
```
r-vieira, line 146: we need to differentiate empty result set from errors to the end-user. Let's discuss after this cycle.

src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -168,34 +184,126 @@ <h5 id="offcanvas1Label"></h5>
...
+        } else if (!hasValue && isLink) {
+            $(`${id} > a`).text('')
+            $(id).prev().hide()
+        } else if (hasValue && !isDate && !isLink) {
+            $(id).prev().show()
+            $(id).text(d)
+        } else if (!hasValue && !isDate && !isLink) {
+            $(id).text('')
+            $(id).prev().hide()
+        }
```
r-vieira, line 305: do we need a fallback option in case none of the previous conditions were met?

r-vieira (approved):
looks good. see inline comments.

PR #438, opened 2024-01-11
Phenotype report submission flow

vlagrassa:
Hook up phenotype table trait selector to trait report submission flow.

Highlights:
- Phenotype db modal can be used to start an analysis or select a trait
  - Accepts Jinja var `access_link` which can be used to customize button behavior
  - HTML table stores a few params in HTML dataset, which can then populate the selectors.
    - TODO: Merge this with the fields tracked by @orzut(?)
- Submission:
  - The "Next" button actually submits the traits
  - Validates the expected number of traits & that species match.
    - TODO: Think about filtering db popup based on species of other trait?
  - Errors are flashed without reloading the page, so we don't clear the user's trait selection(s)
- Trait identification:
  - Traits are now uniquely identified by dataset + name. Even if the CaeNDR name is unique, in the case of traits from a bulk file, we want to know which file they came from.
    - Discuss: Should we use dataset or TraitFile unique ID?
  - In line with ^^, the trait names are stored in the `PhenotypeReport` along with the TraitFile unique ID.  This identifies traits from bulk files (i.e. Zhang traits), and has the added bonus of letting us load the `my-reports` and `all-reports` pages without spamming the datastore for trait display names
    - TODO: What happens if trait file name and phenotype report name mismatch? Currently don't check, although this is in line with what we do for e.g. `Container` fields in `Report` objects
  - Mediate between datastore, SQL, and raw dataset + trait name representations with a new bare-bones `Trait` class.  It's not overly complicated, but the fact that it's useful at all might suggest our current SQL + DS + Google Sheets strategy is a bit bloated... alternatively, maybe it would be useful to have e.g. a `Strain` or `Isotype` class that does sthg similar?
- Miscellaneous:
  - Fixed all spellings of "Zhang"
  - Merged single- and double-trait selection pages down to one HTML file. They were already one view function with two endpoints. This way, only have to implement selection / submission code once
  - Removed a few vestigial bits of code (traits API query, Phenotype FlaskForm)
  - Various minor bug-fixes


r-vieira (approved):
looks great!

PR #439, opened 2024-01-12
Phenotype Report Labels

vlagrassa:
Start implementing better tooltips in line with 1/11/24 meeting.  For styling, I've added config variables that control the elements we might want to change (opacity, etc), to make it easier on @natalieroman in case we want to tweak the tooltips later.

Sample tooltips, as they'll look after this PR:
Double-trait Scatterplot Tooltip Example
Single-trait Ranked Barplot Tooltip Example

Highlights:
- Updates labels
  - Replaces trait names with `x = ...` and `y = ...`
  - Removes the word `Strain: ` and just prints the strain name directly.  We can definitely revert this change.
- Adds config parameters for scatterplot dot opacity
  - `opacity` controls the default opacity
  - `opacity_hover` controls the opacity on mouseover
  - For now, I've set these to 0.75 and 1, mostly to demo how the variables work. We'll definitely want to tweak these values.
- Cleans up spacing
  - Text is now vertically centered :)
  - Adds a bit of padding to the body elements.  We can definitely revert this change / tweak this padding.

---

Note: If you'd like to look at the tooltip for styling without having to mouse over a dot every time, the following HTML will replicate the example tooltips shown above:
```
<div class="tooltip" style="position: relative; opacity: 1; width: fit-content;">
    <p class="tooltip-head"><b>NIC258</b></p>
    <p class="tooltip-body">x = 8.540704986943403</p>
    <p class="tooltip-body">y = 1.675780911513509</p>
</div>

<div class="tooltip" style="position: relative; opacity: 1; width: fit-content;">
    <p class="tooltip-head"><b>ECA2583</b></p>
    <p class="tooltip-body">Value = -35.84013654</p>
</div>
```

### DIFF
--- a/src/modules/site-v2/base/static/js/plot.js
+++ b/src/modules/site-v2/base/static/js/plot.js
@@ -160,6 +160,7 @@ function render_scatterplot_histograms(container_selector, data, config={}) {
 
     // Create a template function for the tooltip
     // By default, show label & value for each axis
+    const tooltip_id       = config['tooltip_id']       || null;
     const tooltip_template = config['tooltip_template'] || ((d, labels) => `
       <p class="tooltip-body">${labels[0]}: ${d[0]}</p>
       <p class="tooltip-body">${labels[1]}: ${d[1]}</p>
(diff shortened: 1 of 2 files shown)


r-vieira (approved):
looks good!

PR #441, opened 2024-01-24
Phenotype report data handling

vlagrassa:


src/modules/site-v2/base/static/js/plot.js
```diff
@@ -54,7 +54,15 @@ function add_histogram_along_axis(d, axis, data, target, config={}) {
...
   const graph_height = config['height']   || 60;
   const color        = config['color']    || 'black';
-  const num_bins     = config['num_bins'] || ((axis.ticks().length - 1) * (config['bins_per_tick'] || 1));
+
+  // Compute the approximate number of bins there should be
+  let num_bins;
+  if (config['bins_per_tick']) {
+    num_bins = (axis.ticks().length - 1) * config['bins_per_tick']
+  } else {
+    const bin_width = config['bin_width'] || ((axis.ticks().length - 1) * 2)
```
r-vieira, line 63: do we worry about the scenario where `config['bin_width'] === 0` is what is intended? In the case, 0 || N will yield to N.

src/modules/site-v2/templates/tools/phenotype_database/report.html
```diff
@@ -38,20 +38,27 @@
         <div class="col col-md-6 col-xl-3 text-bg-light mb-5 p-5 rounded-2 shadow-sm">
             <h2 class="h3 text-center mb-5">Trait Information</h2>
 
+            {%- if data.num_traits > 1 %}
```
r-vieira, line 41: do we need a 'nice' fallback message in case something went wrong and num_traits === 0?

src/pkg/caendr/caendr/utils/data.py
```diff
@@ -197,3 +198,13 @@ def keyset_intersection(*dicts):
...
   return overlap
+
+
+def center_and_scale_data(data):
+  '''
+    Mean-center an array of data and scale it by its standard deviation.
+  '''
+  data   = np.array(data)
+  mean   = np.mean(data)
+  stddev = np.std(data)
```
r-vieira, line 209: since `np.std` can be zero, guard against divide-by-zero in the next line.

r-vieira (approved):
looks great -- see inline comments.

PR #442, opened 2024-01-24
Feature/phenotype table

orzut:
- Added pagination for non-bulk traits
- Fixed the handling of trait categories both when parsing the trait files and rendering in the front-end
- Displaying tags/categories in the filter (but currently they don't work, working on that)

[UPDATE]
- Added functionalities for Search and Filter for non-bulk traits table
- Updated pagination (handling with POST request)


src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -67,21 +67,34 @@ def results_columns():
 def phenotype_database():
   """
     Phenotype Database table (non-bulk)
-  """
+  """  
   # Get the list of traits for non-bulk files
   try:
-    traits_non_bulk = query_phenotype_metadata()
+    page = request.args.get('page', 1, type=int)
```
r-vieira: this is good -- ensure that we are indexing/offsetting from a base `0` or base `1`. In other words, is our first page `0` or page `1`?

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -67,21 +67,34 @@ def results_columns():
...
+    per_page = 10
```
r-vieira: ~~a more official name could be `pagesize`.~~

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -67,21 +67,34 @@ def results_columns():
...
+    query = query_phenotype_metadata()
+
+    # Get the list of unique tags
+    tags = [ tr.tags.split(', ') for tr in query if tr.tags ]
```
r-vieira, line 78: hum....this is a quintessential strings challenge -- what happens if one of the single tags has an inline comma and white space, e.g.: json.stringify(["foobar, foobar", "other"])...? We might need to future-proof it for that scenario.

src/modules/site-v2/base/views/tools/phenotype_database.py
r-vieira: but the API here expects `per_page`, so let's keep it as is.

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -97,9 +110,9 @@ def get_zhang_traits_json():
     draw = request.args.get('draw', type=int)
     start = request.args.get('start', type=int)
     length = request.args.get('length', type=int)
-    search_value = request.args.get('search[value]', '').lower()
+    search_value = bleach.clean(request.args.get('search[value]', '')).lower()
```
r-vieira, line 147: quick note on the guard -- if the request is invalid, e.g.: the `search[value]` doesn't contain a valid value to proceed, reject the request early on.

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -67,21 +67,34 @@ def results_columns():
...
   try:
-    traits_non_bulk = query_phenotype_metadata()
+    page = request.args.get('page', 1, type=int)
+    per_page = 10
+    query = query_phenotype_metadata()
+
+    # Get the list of unique tags
+    tags = [ tr.tags.split(', ') for tr in query if tr.tags ]
+    tags_list = [tg for tr_tag in tags for tg in tr_tag]
+    unique_tags = list(set(tags_list))
```
vlagrassa, line 80: Love the use of `set` and `list` to handle duplicates! :)

Do we need to sort the list of tags before rendering, e.g. alphabetically?  (If not, or if you already do that elsewhere, discard this msg)

src/modules/site-v2/base/views/tools/phenotype_database.py
vlagrassa: Should the `per_page` value be stored as part of a config somewhere?  I'm actually not sure what the best place is, since you could make an argument it's a "front-end" value or a "back-end" value... maybe a question for @r-vieira?

(For now I think a constant val in-line is fine.)

vlagrassa (approved):
Looks good to me! :)

r-vieira (approved):
looks good

PR #444, opened 2024-01-26
Bump `pandas` to `1.1.3`

vlagrassa:
Tried using the fix in the previous version of this branch, but still got a build error.  Bumping `1.0.5` to `1.1.3` in `setup.py` fixed it for me.

### DIFF
--- a/src/modules/db_operations/Dockerfile
+++ b/src/modules/db_operations/Dockerfile
@@ -7,10 +7,6 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
       build-essential \
       libz-dev \
-      python3 \
-      python3-dev \
-      python3-venv \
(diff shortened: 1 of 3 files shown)


vlagrassa:
Note: pandas version is still pinned to `1.0.5` in the site-v2 `requirements.txt` file.  As far as I can tell, we can still build the container / run the site with this config, so I haven't changed it -- should we?

vlagrassa:
I stand corrected, this does in fact break the container.

r-vieira (approved):
As a next step, let's add a guard in the docker file that ensures that the python packages are working as expected. This could be a simple call to a `numpy` or `pandas` call that will only run if all the packages (and version dependencies) are working correctly. This will prevent Docker from building broken images. Let's discuss further offline.

PR #445, opened 2024-01-29
Build phenotype tables in dev environment

vlagrassa:
Small fixes to get phenotype trait flow (mostly) working in dev.

The current dev site is built from a local branch of mine incorporating the latest feature branches. Once all are merged into `development`, I will re-build from that branch and push to dev/QA for testing.

Important changes:
- Standardizes the output files in `launch.json` to all use `temp/local.db`.  This can be changed locally to use a different file
- Renames the `species` field in the `PhenotypeMetadata` SQL table to `species_name`, to match the other tables. We query on this column more generally when deciding which rows to drop, so it's easier to make them match now than trying to rebuild the table later. Unfortunately, this might mean feature branches will have to be refactored once they're merged into `dev` -- I'm happy to take this on.
- Adds a temporary fix to the trait submission pipeline which performs the initial CaeNDR trait paginated query. Clicking to a new page in the database popup will exit the pipeline and return to the database page. I've discussed w @orzut and she'll work on a fix soon; in the meantime, as long as you choose a trait from the first page (or a Zhang trait from any page), the submission pipeline should work in `dev`


src/modules/db_operations/operations.py
```diff
@@ -132,6 +135,23 @@ def drop_and_populate_phenotype_metadata(app, db, species, reload_files=True):
   logger.info("Loading phenotypes...")
   etl_manager.load_tables(PhenotypeMetadata, species_list=species)
 
+def drop_and_populate_phenotypes(app, db, species, reload_files=True):
```
r-vieira, line 138: we should add docstring to the function args and be explicit about inputs and side-effects (dropped tables, etc).

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -183,12 +183,25 @@ def submit_traits():
     initial_trait = None
 
   # Get the full list of traits for non-bulk files
+  # NOTE: Temporary solution, will use DataTables to request this from a different endpoint
+  page = request.args.get('page', 1, type=int)
+  per_page = 10
   try:
-    trait_list = query_phenotype_metadata()
+    query = query_phenotype_metadata()
+    pagination = query.paginate(page=page, per_page=per_page)
```
r-vieira, line 238: consider adding a reminder note about the #paginate() index starting at `1` instead of `0` in flaskalchemy.

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -183,12 +183,25 @@ def submit_traits():
...
   except Exception as ex:
     logger.error(f'Failed to retrieve the list of traits: {ex}')
     abort(500)
 
+  # Get the list of unique tags
+  # NOTE: Temporary solution, should pull from the same endpoint as above
+  try:
+    tags = [ tr.tags.split(', ') for tr in query if tr.tags ]
```
r-vieira, line 246: let's touch base on whether tags need an internal representation vs being string literals. The issue with string literals is that if PI wants to rename them, we have to retrofit & update many records. In other words, Tag(id=tag_machine_name_foo_bar, label="Foo Bar")

src/pkg/caendr/caendr/api/phenotype.py
```diff
@@ -28,7 +28,7 @@ def query_phenotype_metadata(is_bulk_file=False, include_values=False, species:
     # Query by species
     if species is not None:
       if species in Species.all().keys():
```
r-vieira, line 30: consider renaming `species` to `species_names` for semantic alignment with the next line

src/pkg/caendr/caendr/models/run/gcp.py
```diff
@@ -158,7 +158,7 @@ def split_job_name(cls, kind, job_name) -> (str, str):
     # Loop through valid database operations
     for op in DbOp:
       if data_id == make_dns_name_safe(op.name):
-        return _kind, op.name
+        return make_dns_name_safe(kind), op.name
```
r-vieira, line 161: are we checking for overlapping values returned by make_dns_name_safe()? For instance, if someone submits `A-A` and someone else submits `A_A` will these map to the same output value?

src/pkg/caendr/caendr/models/sql/phenotype_metadata.py
```diff
@@ -11,7 +11,7 @@ class PhenotypeMetadata(DictSerializable, db.Model):
 
   trait_name_caendr = db.Column(db.String(), unique=True, primary_key=True)
   trait_name_user = db.Column(db.String(), nullable=True)
-  species = db.Column(db.String())
+  species_name = db.Column(db.String())
```
r-vieira, line 14: Note -- when deploying this code to production, we need clear path to do the migration (renaming of column name), and ensuring it won't break. Test in lower envs with existing data.

src/pkg/caendr/caendr/services/sql/etl/phenotype_db.py
```diff
@@ -47,6 +47,11 @@ def parse_phenotypedb_traits_data(species: Species, **files: LocalDatastoreFile)
           trait_name = header
           trait_value = row[column_header_map[header]]
 
+          # NOTE: If no trait name is provided within the file, this falls back to the entity metadata value
+          #       This might not be necessary, if we enforce a stricter file format (trait name inside file)
+          if trait_name == 'trait_name':
```
r-vieira, line 52: should we rename `trait_name` to `trait_name_key`? E.g.:

``` 
if trait_name_key === 'trait_name'
```

src/pkg/caendr/caendr/services/sql/etl/phenotype_db.py
```diff
@@ -47,6 +47,11 @@ def parse_phenotypedb_traits_data(species: Species, **files: LocalDatastoreFile)
...
+            trait_name = file_path.metadata['trait_name_caendr']
+
           # Skip the rows with "NA" trait values
           if trait_value == 'NA':
```
r-vieira, line 56: side note -- do we trust the `trait_value` to be consistently "cased", i.e.: in our input data, are we normalizing all cases of `na/Na/nA/NA` to the same?

r-vieira (approved):
Great -- see inline comments.

PR #447, opened 2024-02-02
Adding EULA confirmation checkbox

orzut:
Few things to note:
 - I converted ".docx" file to ".md" to render it as a markdown.
 - Both files (.docx and .md) are uploaded to the project's private buckets in GCP. If there's a better way to store them, I'm happy to change it.

### DIFF
--- a/src/modules/site-v2/base/views/strains.py
+++ b/src/modules/site-v2/base/views/strains.py
@@ -45,12 +45,15 @@ from caendr.utils.env import get_env_var
 from caendr.services.email import send_email, ORDER_SUBMISSION_EMAIL_TEMPLATE
 from caendr.services.cloud.sheets import add_to_order_ws, lookup_order
 from caendr.services.cloud.secret import get_secret
+from caendr.services.cloud.storage import get_blob
 
 MODULE_SITE_CART_COOKIE_NAME = get_env_var('MODULE_SITE_CART_COOKIE_NAME')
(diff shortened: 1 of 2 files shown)


src/modules/site-v2/base/views/strains.py
```diff
@@ -324,7 +327,8 @@ def order_page_index():
     'title': "Order Summary",
     'cartItems': cartItems,
     'totalPrice': totalPrice,
-    'form': form
+    'form': form,
+    'EULA_url': get_blob(MODULE_SITE_BUCKET_PRIVATE_NAME, 'EULA_for_CaeNDR_website_Dr_Andersen_1-25-24.md').public_url
```
r-vieira, line 331: we need to normalize all json payloads format -- consider using lowercase with underscores (🐍 case)

r-vieira, line 331: consider moving this value ('EULA_for_CaeNDR_website_Dr_Andersen_1-25-24.md') to the `.env` so that we can swap without having to re-deploy code. With the EULAs, I think everytime there is a change to the text of the EULA, we need to preserve the older ones and timestamp when new ones were deployed. This is the case where someone agrees to a EULA from 6 months ago, and today's EULA has changed. The 'contract' is that the user agreed not to today's EULA, but rather to the one from 6 months ago -- therefor, we should preserve historical record of all EULAs moving forward.

r-vieira: and as a side effect, when we `track` that the user agreed to the EULA, we should reference *which* EULA version they agreed to.

r-vieira (approved):
looks good -- see inline comments

PR #448, opened 2024-02-02
Bugfix: Phenotype table search

vlagrassa:
Credit to @orzut for fixing the search query with `ilike`.

Marking as draft until I can make sure this works in the submission flow table.  Depends on #442, so that one should be merged first.


src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -62,28 +63,64 @@ def results_columns():
 # Main Endpoint
 #
 
-@phenotype_database_bp.route('')
+@phenotype_database_bp.route('', methods=['GET', 'POST'])
```
r-vieira, line 66: we should confirm if having `cache` for `POST` requests will introduce any side effects, especially if the cache behavior is not looking into the payload of the POST.

r-vieira (approved):
looks good

src/modules/site-v2/base/views/tools/phenotype_database.py
vlagrassa: Turns out this is a really obtuse question to try and answer... none of the docs will just explain how `memoize` interacts with a `POST` request ;-;

Found the following on manually setting a cache key, but it's for `cached` rather than `memoize`:
https://flask-caching.readthedocs.io/en/latest/#make-custom-cache-key

In this particular case, our concern would be whether the `memoize` decorator properly accounts for the URL variables in the query string.  We need to re-run the function if the query params are different (e.g. we want different results for page 1 vs page 2), but it can/should use the cached value if all the query params are the same (e.g. an empty search, page 2 will always(?) be the same traits).

The function does return different pages / queries properly, so it's definitely not over-caching.  But this does make me wonder, how would we tell if it's even caching the results at all?  I.e. if every request is slightly different, maybe `memoize` never triggers?

I wonder if it's worth splitting the `POST` behavior out to a separate function (either a view function in the API, or a utility function) that takes all the query params as explicit arguments, then caches / memoizes based on that.  We might also need to use the `cached` decorator instead, using a custom cache key that accounts for all the query params.

PR #449, opened 2024-02-05
Hotfix: Viewing reports for non-admin users

vlagrassa:
If two `User` objects represent the same datastore entry, they are now explicitly considered equal.

This caused a bug for non-admin users trying to access one of their reports -- since the equality check would fail, they would be denied access.

Using a new test account, I've confirmed this bug did not make it to production.


r-vieira (approved):
good

PR #450, opened 2024-02-05
Adding links to Categories in the table

orzut:


src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -197,13 +192,29 @@ <h5 id="offcanvas1Label"></h5>
             sendPostRequest(data)
         })
 
+        $(document).on('click', '[id^="tagLink"]', function() {
```
r-vieira, line 195: this selector is a bit too wide (`document`), try to find the closest parent to where the tag elements are.

r-vieira (approved):
looks good. see inline comment

PR #451, opened 2024-02-06
Feature/phenotype report page

natalieroman:
Made charts responsive and added buttons for the download SVG and download TSV options (commented out for now).

### DIFF
--- a/src/modules/site-v2/base/static/js/plot.js
+++ b/src/modules/site-v2/base/static/js/plot.js
@@ -181,7 +181,7 @@ function render_scatterplot_histograms(container_selector, data, config={}) {
       //  .attr('height', height + margin.top + margin.bottom + hist_height)
       .attr('width', "100%")
       .attr('height', "100%")
-      .attr("viewBox", `-70 0 900 900`)
+      .attr("viewBox", `-25 -50 900 1000`)
       .attr('preserveAspectRatio','xMinYMin')
 
     // Add a graph element for the scatterplot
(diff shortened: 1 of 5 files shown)


src/modules/site-v2/base/static/js/plot.js
```diff
@@ -161,7 +161,7 @@ function render_scatterplot_histograms(container_selector, data, config={}) {
     const height = (config['height'] || 400) - margin.top - margin.bottom;
 
     // Read values from config, filling in default values when not supplied
-    const circle_radius = config['circle_radius'] || 4;
+    const circle_radius = config['circle_radius'] || 8;
```
r-vieira, line 164: consider using `??` for default values instead of `||` -- this avoids an edge case where the actual value retrieved is zero (Valid), but `||` jumps to the default value instead.  

Example `||`:
```
config['circle_radius'] = undefined; // this value is not defined.
const circle_radius = config['circle_radius'] || 8;  // will yield 8, which is correct.

config['circle_radius'] = 0;  
const circle_radius = config['circle_radius'] || 8;  // will yield 8 (!) instead of zero, which is not correct.
```

Alternative `??`:
```
config['circle_radius'] = undefined; // the value for the key `circle_radius` in config is not set (undefined)
const circle_radius = config['circle_radius'] ?? 8;  // will yield 8, which is correct.

config['circle_radius'] = 0; 
const circle_radius = config['circle_radius'] ?? 8;  // will yield 0, which is correct
```

src/modules/site-v2/base/static/js/plot.js
```diff
@@ -177,8 +177,12 @@ function render_scatterplot_histograms(container_selector, data, config={}) {
 
     // Create the SVG object for the full graphic (scatterplot + histograms + margins)
     const svg = d3.select(container_selector).append('svg')
-      .attr('width',  width  + margin.left + margin.right + hist_height)
-      .attr('height', height + margin.top + margin.bottom + hist_height)
+      //  .attr('width',  width  + margin.left + margin.right + hist_height)
```
r-vieira, line 180: comments on code are good, but commented out code should be removed.

r-vieira (approved):
looks good -- see inline comment.

PR #455, opened 2024-02-07
Speeding up the metadata retrieval for non-bulk traits

orzut:


r-vieira (approved):
approved

src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -163,19 +163,25 @@ def get_zhang_traits_json():
...
+  if trait_name:
+    try:
+      trait = get_trait(trait_name).to_json_with_values()
+      return jsonify(trait)
+
+    except Exception as ex:
+      msg = f'Failed to retrieve metadata for trait {trait_name}'
+      logger.error({msg: ex})
+      
+  return jsonify({ 'message': msg }), 404
```
r-vieira, line 184: if we are returning an error, besides the `404` statusCode`, we should standardize sending a key `error` in the response object. E.g.: `{'error': msg} `

r-vieira (approved):

PR #456, opened 2024-02-07
Minor fixes for phenotype report submission

vlagrassa:
NOTE: Changes login behavior to now track full return path (i.e. including URL vars).  I think this should be fine, I just don't know if it was an intentional choice not to pass these originally.


r-vieira (approved):
looks good

PR #460, opened 2024-02-09
Fix: Indel Finder empty results

vlagrassa:
Quick fix that replaces a `None` result with an empty dict and makes sure to use `.get()` whenever accessing it.

Now also aborts with an error if trying to download the result as a file.


r-vieira (approved):
approved

PR #461, opened 2024-02-09
FASTA File Environment Variables

vlagrassa:
Move the FASTA file name template & extensions to environment variables.  This is (a) good practice and (b) lets us easily change which file we want to use in different environments, while still using a single public bucket.

Drastically simplifies how we handle FASTA files (see d7869f5e92b26e35f2a78008c623cc8dcf855fe0 for details).


r-vieira (approved):
looks good

PR #462, opened 2024-02-09
Phenotype DB feature flag

vlagrassa:
Adds `PHENOTYPE_DB_ENABLED` environment variable following discussion with @r-vieira.  If set to False, hides all phenotype database pages / URLs unless the user is an admin.

Note that this also hides the page links from the Tools page & the navbar.  It looks like we don't have it in the footer yet, but when we add it, we should add a feature flag check there too.

Note: Since the blueprint still exists & is added to the app (so that admins can view it), the breadcrumb will still show on the 404 page (see screenshots below).  If we think this is a problem, I can look into hiding it.

Screenshot 2024-02-09 at 4 24 35 PM
Screenshot 2024-02-09 at 4 25 29 PM


natalieroman (approved):

env/development/global.env
```diff
@@ -153,6 +153,9 @@ HERITABILITY_CONTAINER_VERSION=v0.04
 HERITABILITY_TASK_QUEUE_NAME=caendr-h2calc
 HERITABILITY_EXAMPLE_FILE=data/heritability_example_{SPECIES}.tsv
 
+PHENOTYPE_DB_ENABLED=True
```
r-vieira, line 156: consider adding a comment with the expected format for the values (e.g.: literal python bool converted to string).

r-vieira (approved):
looks good

PR #463, opened 2024-02-09
Fix/phenotype table updates

orzut:


src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -171,17 +171,17 @@ def get_traits_json():
...
 
   if trait_name:
     try:
       trait = get_trait(trait_name).to_json_with_values()
       return jsonify(trait)
 
     except Exception as ex:
-      msg = f'Failed to retrieve metadata for trait {trait_name}'
-      logger.error({msg: ex})
+      logger.error({err_msg: ex})
```
r-vieira, line 182: I thought of mentioning this here since the behavior is different than `js`. The keys for python dictionaries are an `expression` type that 'evaluate' to the `key` to be used -- unless they are under quotes, or are a symbol/etc. For instance:
```python
 >>> a = 'key'
>>> b = { a: 42}
>>> b
{'key': 42}
```

r-vieira: with that in mind, consider having a key `error` in the resulting object, and concatenate the exception message to it.

src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -256,28 +264,30 @@ <h5 id="offcanvas1Label"></h5>
             }).get()
         }
         const data = {csrf_token: $('#csrf_token').val(), search_val: $("#traitsSearch").val(), selected_tags: selectedTags}
-        sendPostRequest(data)
+        await sendPostRequest(data)
```
r-vieira, line 267: if possible, avoid mixing async/await with promises.  Consider: 
```js
try {
  const res = await sendPostRequest(data);
  const resp = handlePostRequest(res); // if this is async, use await.
  return resp; 
} catch (err) {
  console.error(err);
  flash_message(`...`);
}
```

src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -256,28 +264,30 @@ <h5 id="offcanvas1Label"></h5>
...
+            .then( resp => handlePostRequest(resp))
+            .catch( e => {
+                console.error(e)
+                flash_message(`Failed to retrieve the list of traits: ${e.statusText}`)
+            })
     }
 
     function sendPostRequest(data) {
-                $.ajax({
+        return new Promise((resolve, reject) => {
```
r-vieira, line 276: consider using the HTML5 `fetch() with async/await` instead of using promises.

src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -378,7 +389,20 @@ <h3 class="h5 text-dark">Full Description</h3>
             search_val: $("#traitsSearch").val(),
             selected_tags: selectedTags
         }
-        sendPostRequest(data)
+        await sendPostRequest(data)
```
r-vieira, line 392: consider rewriting from `promise` style to `async/await` style.

r-vieira (approved):
looks good -- see inline comments.

PR #464, opened 2024-02-12
Adding display names for traits

orzut:


src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -291,15 +291,19 @@ <h5 id="offcanvas1Label"></h5>
...
+
             let html = `
                     <tr class="py-3 dislayedRow">
                         <td>
                             <a href="#" class="text-dark fw-normal" data-bs-toggle="collapse"
                                 data-bs-target="#collapse-${i+1}" aria-expanded="false" aria-controls="collapse-${i+1}"><i
                                     class="bi bi-chevron-right" aria-hidden="true"></i><em>${speciesName}</em></a>
                         </td>
-                        <td id="accordionHeading-${i+1}"><strong>${arr[i].trait_name_caendr}</strong></td>
+                        <td id="accordionHeading-${i+1}">
```
vlagrassa, line 302: One thought here -- if a trait is missing the display name 1 field, but it does have a display name 2 or 3, then this might look odd.  Might consider checking for display name 1 in an if/else block.

(Of course, this should never happen, but you never know)

vlagrassa (approved):

src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -291,15 +291,19 @@ <h5 id="offcanvas1Label"></h5>
                     tagsHTML += `<div class="mb-3"><a role="button" class="btn btn-link" id="tagLink-${j+1}" data-value="${tags[j]}">${tags[j].replaceAll('/', ' • ')}</a></div>`
                 }
             }
-            
+
```
r-vieira, line 294: should this be a `partial` jinja template instead?

r-vieira (approved):
looks good. See inline comments.

src/modules/site-v2/templates/_includes/phenotype-database-table.html
orzut: As we're fetching data on the client-side I was only able to update html with Javascript. But I'll be happy to implement it with Jinja template if there's a way. Please let me know.

PR #467, opened 2024-02-19
Fix DB Stats Page

vlagrassa:
Currently breaking in QA/dev because there's no error handling for the deprecated Homologs table. (Not breaking in prod, where that table still exists.)

### DIFF
--- a/src/pkg/caendr/caendr/services/database_operation.py
+++ b/src/pkg/caendr/caendr/services/database_operation.py
@@ -6,7 +6,7 @@ from caendr.utils.env                 import get_env_var
 from caendr.models.datastore          import DatabaseOperation
 from caendr.models.sql                import DbOp, ALL_SQL_TABLES
 from caendr.services.cloud.datastore  import get_ds_entity, query_ds_entities
-from caendr.services.cloud.postgresql import rollback_on_error
+from caendr.services.cloud.postgresql import rollback_on_error_handler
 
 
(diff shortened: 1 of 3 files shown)


vlagrassa:
Quick note: This version of the row count algorithm gives slightly higher values than the existing one for some tables.  

Comparison:

| table | old count | new count |
| --- | --- | --- |
| strain | 3,905 | 3,905
| wormbase_gene | 1,798,144 | 1,798,446 |
| wormbase_gene_summary | 90,342 | 90,342 |
| strain_annotated_variants | 21,036,834 | 23,000,000 |
| phenotype_db | 4,898,332 | 5,319,812 |
| phenotype_metadata | 25,952 | 25,952 |

src/pkg/caendr/caendr/services/database_operation.py
```diff
@@ -16,67 +15,76 @@
...
   op = get_ds_entity(DatabaseOperation.kind, op_id)
   logger.debug(op)
   return op
 
+
+#
+# Computing SQL Table Stats
+#
+
+@rollback_on_error
```
r-vieira, line 40: if we are reading from DB and there are no side-effects (changes being written to DB), we might not need transaction support/rollback

r-vieira (approved):
looks good.

src/pkg/caendr/caendr/services/database_operation.py
vlagrassa: Possibly... I ran into trouble with Orzu with the trait requests though, since it turns out the `pagination` call can create an invalid transaction even though it's theoretically just reading data.  I lean towards better safe than sorry, unless we have a good reason to remove it.

PR #468, opened 2024-02-19
Phenotype Report Downloads

vlagrassa:
Adds support for downloading reports as TSV and SVG.

Notes:
- I wasn't sure the best way to handle the SVG download for the single trait report page, since there are two SVGs. For now, it downloads them both as individual files one after the other. Open to changing this if there's a better way to do it (combining into one file, etc.)
- Neither TSV nor SVG format currently include the correlation or p-values. Will we need those in a download?

Questions:
- Should we add SVG downloads for the other tools that produce D3 charts? Namely Indel Finder and Heritability Calculator.


vlagrassa:
Updates:
- SVG download buttons (merged from dev) work properly
- Added a PDF download button -- grabs trait information, chart(s), and p-value / correlation / mean-centering message. Let me know if this is what we're looking for

natalieroman (approved):
Looks good!

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -92,10 +92,7 @@ function force_download(e) {
       // If everything succeeds, all future clicks will use the modified href
       // in the existing anchor element. If something fails, all clicks should
       // restart the blob download process, which is is acceptable.
-      const a = document.createElement('a');
-      a.href = el.href;
-      a.download = el.download || 'download';
-      a.click();
+      download_from_href(el.href, el.download || 'download');
```
r-vieira, line 95: consider renaming `download_from_url()`, since semantically, it could come from a non `href` prop.

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -108,6 +105,23 @@ function create_node(html) {
 }
 
 
+function save_svg(selector, filename=null) {
```
r-vieira, line 108: consider setting the default value for filename as a default argument, eg:
`save_svg(selector, filename = 'download.svg') {`

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -108,6 +105,23 @@ function create_node(html) {
...
+  const svg_el = document.querySelector(selector);
+  const data = (new XMLSerializer()).serializeToString(svg_el);
```
r-vieira, line 110: are we catching errors from serialization/blob creation/URL/etc upstream?

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -108,6 +105,23 @@ function create_node(html) {
...
+  const svg_blob = new Blob([data], {type: 'image/svg+xml;charset=utf-8'});
+  const url = URL.createObjectURL(svg_blob);
+  download_from_href(url, filename || 'download.svg');
+}
+
+
+function download_from_href(href, filename) {
```
r-vieira, line 117: consider renaming function to `download_from_url(...){`

src/modules/site-v2/templates/_scripts/utils.js
```diff
@@ -108,6 +105,23 @@ function create_node(html) {
...
+  const a = document.createElement('a');
+  a.href     = href;
+  a.download = filename || 'download';
+  a.click();
```
r-vieira, line 121: should re remove the `a` element from the DOM once the download has been initiated?

src/modules/site-v2/base/static/js/plot.js
```diff
@@ -79,7 +80,7 @@ function add_histogram_along_axis(d, axis, data, target, config={}) {
   const bins = d3.histogram()
     .domain(axis.domain())
     .thresholds(axis.ticks(num_bins))
-    .value(n => n[d])
+    .value(n => n[d + 1])
```
r-vieira, line 83: consider adding a comment about the need for `+1`, or create a named variable that hints at why we need to offset the index.

r-vieira (approved):
looks great. See inline comments.

r-vieira:
> * I wasn't sure the best way to handle the SVG download for the single trait report page, since there are two SVGs. For now, it downloads them both as individual files one after the other. Open to changing this if there's a better way to do it (combining into one file, etc.)

We might run into some browsers blocking multiple downloads simultaneously (or add a banner/require end-user approval to start the download). Maybe we could add download actions for both and let the end-user download each individually.

r-vieira:
> * Neither TSV nor SVG format currently include the correlation or p-values. Will we need those in a download?

let's bring this up to Erik.

r-vieira:
> * Should we add SVG downloads for the other tools that produce D3 charts? Namely Indel Finder and Heritability Calculator.

excellent question! Let's discuss with Erik.

PR #469, opened 2024-02-20
Fixes for the previous PRs

orzut:


src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -270,27 +270,32 @@ <h1 class="h5" id="offcanvas1Label"></h5>
...
-                type: "POST",
-                url: "{{ url_for('phenotype_database.phenotype_database') }}",
-                data: JSON.stringify(data),
-                contentType: "application/json",
-                dataType: "json",
-                success: (response) => resolve(response),
-                error: (error)=> reject(error)
-            })
+    async function sendPostRequest(data) {
+        const resp = await fetch("{{ url_for('phenotype_database.phenotype_database') }}", {
```
r-vieira, line 283: double check if `fetch` throws exceptions -- if so, you might need to wrap the call to `await fetch(...)` into a try/catch block.

src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -270,27 +270,32 @@ <h1 class="h5" id="offcanvas1Label"></h5>
...
+            headers: {
+                "Content-Type": "application/json",
+                'X-CSRF-TOKEN': data.csrf_token
+            },
+            body: JSON.stringify(data)
         })
-                
+
+        if (resp.ok) {
+            const respData = await resp.json()
```
r-vieira, line 293: if the server returns html instead of valid json, this call (`await res.json()`) will also throw an exception and will need  to be wrapped into try/catch. You might combine both the `await fetch()` and `await res.json()` int he same try/catch block, if it makes sense.

r-vieira (commented):
approved -- see inline notes about handling exceptions on #fetch() and #json()

PR #470, opened 2024-02-23
Fix/trait name styling

natalieroman:
Improved style/layout so that the new trait name format (3 fields) looks good on analysis and report pages.


r-vieira (approved):
looks good.

PR #472, opened 2024-02-29
Render trait display names everywhere

vlagrassa:
Updates all pages/widgets that deal with traits to use the full trait display names.  This includes:

- Phenotype database table rows
- Phenotype database offcanvas header
- My Report / All Reports table rows
- Trait selectors (single and double) (when creating/submitting a report)
- Reports page sidebar
- Reports page graph axis labels

Notes:
- The `PhenotypeReport` objects cache the display name, so we can render them in the My Reports page without having to ping the database 1~2x for every report. I've added a DbOp that re-computes these cached names, just in case
- For now, the Zhang traits just use the full CaeNDR name as the `display_name_1` field. If we wanted to, we could try to split these names up based on the columns in the sheet.

Screenshot 2024-02-29 at 5 52 26 PM


natalieroman (approved):

src/modules/db_operations/seed_trait_files.py
```diff
@@ -61,9 +61,12 @@ def populate_andersenlab_trait_files():
         'species':           get_field_from_record(record, 'Species', nullable=False),
 
         # About trait
-        'description_short': get_field_from_record(record, 'Short_Description'),
-        'description_long':  get_field_from_record(record, 'Long_Description'),
-        'units':             get_field_from_record(record, 'Units'),
+        'trait_name_display_1': get_field_from_record(record, 'Trait_Name_Display1'),
```
r-vieira, line 99: can we add an `assert()` that all the expected Keys (headers) exist in the sheet as soon as possible and bail out accordingly (show error message about missing keys, case-sensitive keys, etc).

src/modules/site-v2/templates/_scripts/trait_utils.js
```diff
@@ -0,0 +1,44 @@
...
+
+  // Construct the data object, using the optional CSRF token if provided
+  let data = {trait_name};
+  if (csrf_token !== null) data['csrf_token'] = csrf_token;
+
+  // Return the AJAX request as a Promise object
+  return $.ajax({
+      type:        "POST",
+      url:         "{{ url_for('phenotype_database.get_traits_json') }}",
+      data:        JSON.stringify({trait_name}),
```
r-vieira: should we pass in `data` to `data` instead of just {trait_name} ?

src/modules/site-v2/templates/tools/phenotype_database/submit-traits.html
```diff
@@ -130,14 +137,17 @@ <h5 class="modal-title" id="confirmationModalLabel">Thank you</h5>
...
-            }
+            queryTraitByName(trait)
+                // Re-call the function using the trait value instead of the name
+                // Response check prevents infinite recursion
+                .then((response) => {
+                    if (response !== null && typeof response === 'string') throw Error(`Invalid response for trait ${id}: "${response}"`);
+                    select_trait(id, response, dataset);
+                })
+                .fail((err) => {
+                    console.error(error);
```
r-vieira, line 149: should we notify the end-user of the error?

src/modules/site-v2/templates/tools/phenotype_database/submit-traits.html
```diff
@@ -172,20 +182,20 @@ <h5 class="modal-title" id="confirmationModalLabel">Thank you</h5>
...
-        else {
-            $('#tableAccordion tbody>tr').hide().each(function() {
-                console.log(this.dataset.species, species)
-                if (this.dataset.species == species) {
-                    $(this).show()
-                }
-            });
-        }
-    }
+    // function filter_by_species(species) {
```
r-vieira, line 195: we can remove commented code

r-vieira (commented):
looks good. See inline comments.

src/modules/db_operations/seed_trait_files.py
vlagrassa: Added in [011de8d](https://github.com/AndersenLab/CAENDR/pull/472/commits/011de8d8302fa12cb444e9bbbc04391e38e82746) -- check for set of expected keys, and exit the function with an error if any are missing.  This will terminate the db operation with an error.

src/modules/site-v2/templates/_scripts/trait_utils.js
vlagrassa: Oops, good catch! Fixed in [db31f33](https://github.com/AndersenLab/CAENDR/pull/472/commits/db31f3344d804d31f014f6522993a41c32e0c342)

src/modules/site-v2/templates/tools/phenotype_database/submit-traits.html
vlagrassa: Added in [75e9873](https://github.com/AndersenLab/CAENDR/pull/472/commits/75e98735e7c7ebba363679189c0ff72b7eaf1529) -- flashes the error message for the user, like we do elsewhere.

src/modules/db_operations/seed_trait_files.py
```diff
@@ -13,6 +13,31 @@
...
+  'Short_Description',
+  'Long_Description',
+  'Units',
+  'Publication',
+  'Protocol',
+  'Institution',
+  'Captured_By_UserID',
+  'Capture_Date',
+
+  # TODO: How should we handle category tags?
```
r-vieira, line 34: we don't expect these to change too often, but when they do, it would be nice to not have to make code changes to support it.

r-vieira (approved):
looks good.

src/modules/site-v2/templates/tools/phenotype_database/submit-traits.html
```diff
@@ -130,14 +137,27 @@ <h5 class="modal-title" id="confirmationModalLabel">Thank you</h5>
...
+                // Re-call the function using the trait value instead of the name
+                // Response check prevents infinite recursion
+                .then((response) => {
+                    if (response !== null && typeof response === 'string') throw Error(`Invalid response for trait ${id}: "${response}"`);
+                    select_trait(id, response, dataset);
+                })
+                .fail((error) => {
+                    // If an error message was received, display it, otherwise use a fallback message
+                    console.error(error);
+                    const resp = error.responseJSON
```
r-vieira, line 150: if throwing from line `144`, will the property `.responseJSON` be present in the `error` object? 

Also, it might be helpful to document the expected shape of the `error.responseJSON` structure.

src/modules/site-v2/templates/tools/phenotype_database/submit-traits.html
```diff
@@ -130,14 +137,27 @@ <h5 class="modal-title" id="confirmationModalLabel">Thank you</h5>
...
+                    if (resp && resp.message && false) {
```
r-vieira, line 151: can we clarify the intention here? assuming `resp` is object/null, `resp.message` is undefined/string/object then using `false` appears to convolute the intention with mixed types.

r-vieira (commented):
looks good. See inline comments

src/modules/site-v2/templates/tools/phenotype_database/submit-traits.html
vlagrassa: Ah, this would be an API call error, not a JS error from the success block.

As far as I can tell, error objects always have a `responseJSON` property, although I do check in the next line to make sure the object exists (`if (resp ...`)

src/modules/site-v2/templates/tools/phenotype_database/submit-traits.html
vlagrassa: Hm... I think I threw `false` in here because I wanted to keep this structure (which I use elsewhere for error handling) but I only want to flash the default message?  Usually it's `if (resp && resp.message)`, to just check if the `message` field exists.  Definitely weird though, I'll clean it up.

PR #473, opened 2024-03-13
Keep Pub/Sub from deleting itself

vlagrassa:
Set expiration policy for subscription to "never" (`""` in Terraform), overwriting default 31 days.  A subscription with an expiration policy will delete itself if there is no activity for that length of time, so if no one submitted a job for that long, the pipeline would break.

### DIFF
--- a/tf/caendr/modules/api/pipeline_task/pub-sub.tf
+++ b/tf/caendr/modules/api/pipeline_task/pub-sub.tf
@@ -24,7 +24,7 @@ resource "google_pubsub_subscription" "pipeline_task" {
   enable_message_ordering = "false"
 
   expiration_policy {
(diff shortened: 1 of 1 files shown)


tf/caendr/modules/api/pipeline_task/pub-sub.tf
```diff
@@ -24,7 +24,7 @@ resource "google_pubsub_subscription" "pipeline_task" {
   enable_message_ordering = "false"
 
   expiration_policy {
-    ttl = "2678400s"
+    ttl = ""
```
r-vieira, line 27: 👍

r-vieira (commented):
good catch

PR #476, opened 2024-04-22
[Lower priority] Enhancement: Parsing decorators for common Entity lookups & form verification

vlagrassa:
Lower priority -- this improves the logic / readability of a number of job endpoint functions, and standardizes error handling across analogous endpoints (both for different tools and just in general).  Does not strictly speaking add new functionality.

This PR creates a few decorator functions in the new file `src/modules/site-v2/base/utils/view_decorators.py`, designed to streamline the process of looking up / parsing the data that goes into an endpoint function.

New decorators:
- `parse_species` and `parse_species_and_release`: Given a `species_name` string and optional `release_version` string in the URL, lookup the corresponding `Species` and `DatasetRelease` objects, raising the appropriate error if these do not exist or do not match.  Passes the objects into the (wrapped) endpoint function.
- `parse_job_id`: Given a `report_id` string in the URL and an endpoint-specific `JobPipeline` subclass, look up the job for that pipeline with the given ID and retrieve its data.  Raises the appropriate errors if the job cannot be found or if the data does not exist.  Passes the job directly into the (wrapped) endpoint function, along with the data and results if applicable.
- `validate_form`: Given a `FlaskForm` and a request with form data, parse out and validate the form data. Includes retrieving an optional file upload, and handles creation/deletion of that file in local memory.  Passes the parsed form data into the (wrapped) endpoint function.

For more info on any specific function, see the entry in `view_decorators.py`.  For usage, primarily see the view functions for the user-submission tools (Indel Finder, Nemascan, Heritability, Phenotype Analysis).

(Originally started working on this PR a few months ago, then set it aside as work for Phenotype Analysis picked up.  The decorators in this PR are useful for the Analysis tool though, so I've dusted it off and applied the changes to the Analysis tool too.)


src/pkg/caendr/caendr/models/report/gcp_report.py
```diff
@@ -45,6 +45,10 @@ class GCPReport(BucketedReport):
   def _generate_uri(cls, bucket: str, *path: str, schema: BlobURISchema=None):
     return generate_blob_uri(bucket, *path, schema=schema)
 
+  @classmethod
+  def _list_files(cls, bucket: str, *prefix: str, filter=None):
+    return get_blob_list(bucket, *prefix, filter=filter)
```
r-vieira, line 50: how is pagination handled if the number of files (objects) exceeds the default pagesize? (see: https://cloud.google.com/storage/docs/paginate-results)

src/pkg/caendr/caendr/models/report/bucketed_report.py
```diff
@@ -36,6 +36,11 @@ class BucketedReport(Report):
   def _generate_uri(cls, bucket: str, *path: str, schema: BlobURISchema=None):
     pass
 
+  @classmethod
+  @abstractmethod
+  def _list_files(cls, bucket: str, *prefix: str, filter=None):
+    pass
```
r-vieira, line 42: consider raising NotImplementedError in the abstractmethod to ensure subclasses implement it.

src/modules/site-v2/base/utils/view_decorators.py
```diff
@@ -0,0 +1,224 @@
...
+      # Upload input file to server temporarily and add to the list of form fields
+      # TODO: This hardcodes the field name 'file' for a *single* file upload -- generalize whatever file field(s) are present
+      try:
+        with LocalUploadFile(request.files['file'], valid_file_extensions=TOOL_INPUT_DATA_VALID_FILE_EXTENSIONS) as local_file:
+
+          # Pass the objects to the function
+          return f(*args, form_data={**form_data, 'file': local_file}, no_cache=no_cache, **kwargs)
+
+      # If the file upload failed, display an error message
+      except FileUploadError as ex:
```
r-vieira, line 218: do we need to clean up temporary files at this stage in exception handling?

src/modules/site-v2/base/utils/view_decorators.py
```diff
@@ -0,0 +1,224 @@
...
+      # e.g. not enough traits in a Phenotype Analysis report
+      except DataValidationError as ex:
+        logger.error(f'Error fetching {pipeline_class.get_kind()} report {id}: {ex}')
+        flash(ex.msg, 'error')
+        return abort(400, description = ex.msg)
+
+      # General error
+      except Exception as ex:
+        logger.error(f'Error fetching {pipeline_class.get_kind()} report {id}: {ex}')
+        return abort(400, description = 'Something went wrong')
```
r-vieira, line 135: prefer slightly more descriptive error messages that can help us troubleshoot issues down the road, especially to identify at what layer the issue occurred. If we don't want to disclose details on internals, we can come up with a ERROR_CODE set of enums (this would be something we can quickly identify on our end).

src/modules/site-v2/base/utils/auth.py
```diff
@@ -172,3 +172,24 @@ def expired_token_callback(_jwt_header, jwt_data):
...
+    Require a "Bearer" access token in the request.
+  '''
+  def wrapper(fn):
+    @wraps(fn)
+    def decorator(*args, **kwargs):
+
+      # Check for access token in request
+      access_token = request.headers.get('Authorization')
+      if access_token != 'Bearer {}'.format(token):
+        abort(403)
```
r-vieira, line 189: quick note on semantics -- should this be 401 (authn: user didn't provide valid credentials) vs 403 (authz: user provided valid credentials, but is not authorized to do this action).

r-vieira (approved):
great work! see inline comments.

src/modules/site-v2/base/utils/view_decorators.py
vlagrassa: I think so, since we exit the view function here.

(For context, the `LocalUploadFile` context manager cleans up its temporary file on exit.)

src/pkg/caendr/caendr/models/report/bucketed_report.py
vlagrassa: I think this makes sense -- I was tempted to leave it as-is, so we don't break `super()` calls in case of multiple inheritance, but (a) I'm not sure multiple inheritance makes sense here and (b) if we really wanted to do that, this should return a default value like `[]` instead.  Updated in 641a24932affa7adb55bb49735c2c334bbcb5b62.

src/modules/site-v2/base/utils/view_decorators.py
vlagrassa: Yep, that was my concern -- I'm hesitant to abort with an arbitrary Exception message, since it might (and probably should) contain internal details.  The internal ERROR_CODE idea sounds interesting... also wonder if we could attach an "error ID" to certain `abort`ed results that we track in the logs, so we can easily locate them in the GCP console?

src/modules/site-v2/base/utils/auth.py
vlagrassa: Good catch, fixed in 25cfcc10b9f530cd5e69dba966edd5f731790d8c

src/pkg/caendr/caendr/models/report/gcp_report.py
vlagrassa: Looks like our implementation of `get_blob_list` iterates through the result and returns as a list, so it should consume all results across multiple paginated requests:

https://github.com/AndersenLab/CAENDR/blob/643d52dbc679f993fd5abc083885957a5d86568f/src/pkg/caendr/caendr/services/cloud/storage.py#L77-L92

PR #477, opened 2024-04-23
Trait Submission Form

orzut:
- UI + view function for Trait Submission start page
- UI + view functions for Trait Submission Form
- Add a new trait as a new TraitFile object (datastore)
- Add a new trait to PhenotypeMetadata SQL table
- Validate the uploaded file
- Upload the file to GCP buckets
- Parse the file and seed to PhenotypeDatabase SQL table
- Added new fields to both SQL table (id and metadata_id for PhenotypeMetadata and PhenotypeDatabase respectively). @vlagrassa after this PR gets merged we'll need to re-populate phenotype databases.  Also, I think this change will affect other parts of the site's pipeline too. I'm happy to discuss the details.


src/modules/site-v2/base/views/data/data.py
```diff
@@ -57,3 +76,186 @@ def protocols():
     }
     return render_template('data/protocols.html', **params)
 
+
+#
+# Submit Trait
+#
+@data_bp.route('/submit-trait')
```
r-vieira, line 83: For the endpoints, let's normalize around one of the standards, eg: REST:
- /content/traits [GET] list all the traits
- /content/traits [POST] create a new trait
- /content/trait/:id [GET] View single trait
- /content/trait/:id [PUT] Modify single trait
- /content/trait/:id [DELETE] Delete single trait.

Since some of these operations will require an edit/modify view or delete confirmation, consider also these:
- /content/traits/create [GET] form for new trait creation. The form posts to /content/traits [POST]
- /content/traits/:id/edit [GET] form to edit the trait. The form 'puts' to /content/traits/:id [PUT] 
- /content/traits/:id/delete [GET] form to confirm the deletion of the trait. The form posts to /content/traits/:id [DELETE]

src/modules/site-v2/base/views/data/data.py
```diff
@@ -57,3 +76,186 @@ def protocols():
...
+@jwt_required()
+def submit_trait_form():
+  """ Trait Submission Form """
+  form = TraitSubmissionForm()
+  user = get_current_user()
+
+  if hasattr(user, 'username') and not form.username.data:
+    form.username.data = user.username
+
+  # Handle form submission
```
r-vieira, line 105: an architecture suggestion -- borrowing a page from the [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller), see if we can decouple the `controller` role of this function in handling the HTTP request from the data manipulation that we need to apply `model`. In this case, consider creating another function that handles the creation of a trait where it has all the business logic to validate and create the trait artifacts (data store, file uploads, etc). One side benefit is that we could then call the trait function individually for automated tests while mocking the request payload without having to call this function directly.

src/modules/site-v2/base/views/data/data.py
```diff
@@ -57,3 +76,186 @@ def protocols():
...
+  if request.method == 'POST':
```
r-vieira, line 106: if this function *only* handles `POST` requests, consider doing early-return, eg:
```python
if request.method != 'POST':
  # flash_message('Operation not supported') # or something to that effect. 
  # redirect to a good page. 
```

src/modules/site-v2/base/views/data/data.py
```diff
@@ -57,3 +76,186 @@ def protocols():
...
+        tf.set_user(user)
+
+        # Save the TraitFile object to Datastore
+        tf.save()
+
+        # Seed to Phenotype Metadata SQL table
+        new_trait = PhenotypeMetadata()
+        new_trait.add_trait(tf)
+
+      except Exception as ex:
```
r-vieira, line 155: I like that we are handling exceptions at all levels, but I wonder if we need to 'rollback' in case the DataStore succeeds, but the SQL operation fails -- should we remove the DataStore entry in that case? If we keep the entry and the user retries the operation, will that lead to 'ghost' traits in the datastore?

src/modules/site-v2/base/views/data/data.py
```diff
@@ -57,3 +76,186 @@ def protocols():
...
+        logger.error(f'Failed to create a trait file {form.trait_name_user.data}: {ex}')
+        flash('Failed to submit a form. Please try again later.', 'danger')
+        abort(500)
+      
+      # Save file to GCP bucket
+      species_name = Species.get(form.species.data).name
```
r-vieira, line 161: similar 'rollback' concern -- if the file upload fails, do we rollback the previous steps?

src/modules/site-v2/base/views/data/data.py
```diff
@@ -57,3 +76,186 @@ def protocols():
...
+                  'strain_name': row[0],
+                  'trait_value': row[1],
+                  'metadata_id': tf.name
+                }
+                trait_data_list.append(trait_data)
+
+        trait_data = PhenotypeDatabase()
+        trait_data.add_trait_data(trait_data_list)
+
+      except FileUploadError as ex:
```
r-vieira, line 212: besides the `FileUploadError` the code above could fail due to other content validation or syntax issues, we might want to add also an `except Exception as e:` to this try block.

src/modules/site-v2/templates/_includes/breadcrumb.html
```diff
@@ -11,7 +11,7 @@
             {% elif alt_parent_breadcrumb %}
             <li class="breadcrumb-item">{{ alt_parent_breadcrumb["title"] }}</li>
             {% elif tool_alt_parent_breadcrumb %}
-                {% if tool_alt_parent_breadcrumb["title"] == 'Strain Catalog' %}
+                {% if tool_alt_parent_breadcrumb["title"] == 'Strain Catalog' or  tool_alt_parent_breadcrumb["title"] == 'Submit Trait'%}
```
r-vieira, line 14: this code works, but it is brittle if we tweak the wording of the title. Do we have a better internal `id` that we can reference? Something that wouldn't need to change if we tweak the wording of the title.

src/modules/site-v2/templates/_includes/phenotype-database-table.html
```diff
@@ -420,9 +435,13 @@ <h3 class="h5 text-dark">Full Description</h3>
     }
 
     function fillOutModal(trait) {
-        $('#offcanvas1-header-1').text(trait.trait_name_display_1)
-        $('#offcanvas1-header-2').text(trait.trait_name_display_2)
-        $('#offcanvas1-header-3').text(trait.trait_name_display_3)
+        let traitName = ''
```
r-vieira, line 438: minor coding style -- we should keep the `;` at the end of lines.

src/modules/site-v2/templates/data/submit-trait-form.html
```diff
@@ -0,0 +1,185 @@
...
+            $('#file-upload').prop('disabled', !$('#speciesSelect').val() || !$('#file').val())
+        })
+
+        $('#file').on('change', function() {
+            $('#file-content').empty()
+            $('#file-content').parent().addClass('d-none')
+        })
+
+        // Upload the file and display the content
+        $('#file-upload').on('click', function(e) {
```
r-vieira, line 125: one front-end architecture idea (loosely similar to the MVC comment earlier), see if we can decouple the event handling from the action itself; especially when the action is more than a few lines. Consider the following: 

```js

// declare and implement the action as a function 
const handleFileUpload = async (...) => {};

// then use it in the event handler
$('#file-upload').on('click',  [async] (event) => {
  event.preventDefault();
  [await] handleFileUpload(...);
});
```
operators in square brackets are optional, use if needed.

src/modules/site-v2/templates/data/submit-trait-form.html
```diff
@@ -0,0 +1,185 @@
...
+                        if (error.responseJSON && error.responseJSON.message) {
+                            $('.alert').remove()
+                            flash_message(error.responseJSON.message)
+                        }         
+                    }
+                })            
+            }            
+        })
+
+        function display_file_content(file) {
```
r-vieira, line 159: thought -- if briefly squinting at this code, it starts to feel like a....react/vue `component`....:)

src/pkg/caendr/caendr/models/datastore/file_record_entity.py
```diff
@@ -24,6 +24,7 @@ def get_props_set(cls):
     return {
       *super().get_props_set(),
       'filename',
+      'hashed_filename',
```
r-vieira, line 27: when naming things, in this case, prefer most-important-to-least-important; ie. `filename_hash`

src/pkg/caendr/caendr/models/datastore/file_record_entity.py
```diff
@@ -79,6 +80,26 @@ def filename(self, v):
     if not (isinstance(v, str) or v is None):
       raise ValueError(f'Cannot set prop "filename" to "{v}" (type {type(v)}): must be a string')
     return self._set_raw_prop('filename', v)
+  
+  @property
```
r-vieira, line 84: +1 for using @property!

src/pkg/caendr/caendr/models/datastore/trait_file.py
```diff
@@ -74,6 +74,8 @@ def bucket(self):
 
   @property
   def prefix(self):
+    if self.dataset == 'public':
```
r-vieira, line 77: try to avoid string literals for any comparisons used for business logic. Instead, prefer comparing using enumerators: 
```python
class DatasetMode(Enum):
    PUBLIC = 1
    NOT_PUBLIC = 2
    NOT_NOT_PUBLIC = 3
```

Then use it as:
```python
  if self.dataset == DatasetMode.PUBLIC
```

r-vieira (commented):
💪 well done on such massive PR! See inline comments -- i'm happy to elaborate on any of them as needed.

r-vieira (approved):

src/modules/site-v2/base/views/data/data.py
orzut: Thanks for touching the base on this. Just to clarify -- do we need to keep all traits endpoints under the data, like `/data/trait(s)/*` or you think they should under separate route (like you proposed "/content"?

src/modules/site-v2/base/views/data/data.py
orzut: It handles the `GET` request too, which renders the form

src/modules/site-v2/base/views/data/data.py
orzut: That's a good point. I'll implement this.

src/modules/site-v2/base/views/data/data.py
r-vieira: sry, not at all, keeping them under `/data/*` is perfectly fine.

src/modules/site-v2/base/views/data/data.py
r-vieira: this is a good candidate for the `command pattern`: https://www.patterns.dev/vanilla/command-pattern/

(more details -- mainly, the ability to 'undo' commands for a rollback):
https://refactoring.guru/design-patterns/command#:~:text=Command%20is%20a%20behavioral%20design,execution%2C%20and%20support%20undoable%20operations.

src/modules/site-v2/templates/_includes/breadcrumb.html
orzut: We don't have any IDs for titles, but I'll see if I can tweak it with `request.blueprint`, which would be more reliable.

src/modules/site-v2/base/views/data/data.py
orzut: Pushed in the commit c3b030566d24eb320d154dac1f8c1f418ad61c95

src/modules/site-v2/base/views/data/data.py
orzut: Pushed in the commit 2e21cc919bfb4210ba96e2b1e215fba0508dce51

src/modules/site-v2/base/views/data/data.py
orzut: Pushed in the commit b96595d18c0d3fd589ea8793398ff122082faad0

src/modules/site-v2/base/views/data/data.py
orzut: Pushed in the commit 4ef3096c658aadbd507cd252de95181cc723305d

src/modules/site-v2/templates/_includes/breadcrumb.html
orzut: Pushed in the commit 15798ce7eb7b7ad089651dc5731f1e0e03546727

src/modules/site-v2/templates/data/submit-trait-form.html
orzut: Pushed in the commit 640f8f002e1e2a30d087d72f8b3e9f55d1c1d050

src/pkg/caendr/caendr/models/datastore/file_record_entity.py
orzut: Pushed in the commit cfcc1cb3f94f015cf32a978d11ee26caa2409f32

src/pkg/caendr/caendr/models/datastore/trait_file.py
orzut: Pushed in the commit d94bf6173d7cfda13e8224c7badddbdc7ae7ff9b

PR #481, opened 2024-05-06
[Trait Submission Form] Changes for the previous PR

orzut:


src/modules/site-v2/base/views/data/data.py
```diff
@@ -80,7 +71,7 @@ def protocols():
 #
 # Submit Trait
 #
-@data_bp.route('/submit-trait')
+@data_bp.route('/trait/start-submit')
```
r-vieira, line 74: 👍

src/pkg/caendr/caendr/models/datastore/trait_file.py
```diff
@@ -74,9 +79,9 @@ def bucket(self):
 
   @property
   def prefix(self):
-    if self.dataset == 'public':
-      return TokenizedString(join_path('trait_files', self['dataset'], '${SPECIES}', '${USER_ID}'))
-    return TokenizedString(join_path('trait_files', self['dataset'], '${SPECIES}'))
+    if self.dataset == DatasetType.PUBLIC:
```
r-vieira, line 82: should we use `self.dataset_type` instead?

r-vieira (approved):
great job!

PR #485, opened 2024-05-20
Feature/moderation queue

orzut:


src/modules/site-v2/base/utils/trait.py
```diff
@@ -136,7 +136,7 @@ def add_trait(form_data, user):
     logger.error(f'Failed to upload the file data to the database: {ex}')
     return {'message': 'Failed to submit a form. Please try again later.'}, 500
   
-  return {'message': 'Trait submitted successfully.'}, 200
+  return {'message': 'Trait submitted successfully.', 'trait_id': tf.name}, 200
```
r-vieira, line 139: can we unify `trait_id` and `trait_name`?

src/modules/site-v2/base/utils/trait.py
```diff
@@ -168,4 +168,46 @@ def rollback_submission_on_error(trait_id, blob_name=None):
...
+      'publication':          bleach.clean(form_data['publication']),
+    }
+    tf.set_properties(**props_to_update)
+    tf.save()
+  except Exception as ex:
+    logger.error(f'Failed to update the trait file {tf.name}: {ex}')
+    return {'message': 'Failed to update the trait. Please try again later.'}, 500
+  
+  try:
+    trait_sql = get_trait(id)
```
r-vieira, line 198: as a convenience, we should use `trait` without the `_sql` suffix since it is an implementation detail that this function might not need to know about.

r-vieira (approved):
looks good

PR #492, opened 2024-06-12
Hotfix/enable pdb qa

orzut:
- Enabling Phenotype Database tool in QA
- Making categories fetch case insensitive


src/pkg/caendr/caendr/models/sql/phenotype_metadata.py
```diff
@@ -62,7 +62,7 @@ def get_tags(self):
       If no tag set defined, returns None.
     '''
     if self.tags:
-      return { tg.strip() for tg in self.tags.split(',') }
+      return { tg.strip().lower() for tg in self.tags.split(',') }
```
r-vieira, line 65: minor detail -- do we need a space after the comma, eg: `, `?

r-vieira (approved):
looks good! See inline comments

PR #493, opened 2024-06-25
Enhancement/clean up phenotype db

orzut:
Joined PR with @vlagrassa for the latest changes to Phenotype DB and other small changes.


src/modules/site-v2/base/forms/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def _format_form_error_val(form: FlaskForm, val):
+  '''
+    Format a value in a form error dict for printing.
+  '''
+  return " ".join(val)
+
+
+def format_form_errors(form: FlaskForm, format = None):
```
r-vieira, line 22: consider:
```python
def format_form_errors(form: FlaskForm, format = lambda key, val: f'{key}: {val}'):
```

src/modules/site-v2/base/views/data/data.py
```diff
@@ -8,22 +8,28 @@
...
+from base.utils.auth               import jwt_required, get_current_user, user_is_admin, check_feature_flag
+from base.utils.trait              import add_trait, update_trait_metadata
+from base.utils.view_decorators    import parse_trait
+from base.forms                    import TraitSubmissionForm, EmptyForm, format_form_errors
 from constants                     import TOOL_INPUT_DATA_VALID_FILE_EXTENSIONS
 
 MODULE_DB_OPERATIONS_BUCKET_NAME = get_env_var('MODULE_DB_OPERATIONS_BUCKET_NAME')
 UPLOADS_DIR = os.path.join('.', 'uploads')
 
+SITE_OWNER = User.get_ds(get_secret('SITE_OWNER_USER_ID'))
```
r-vieira, line 31: consider adding a graceful fallback if the SITE_OWNER returns `None` (in case the data store is not populated, etc).

src/modules/site-v2/base/views/data/data.py
```diff
@@ -75,17 +81,20 @@ def protocols():
...
 
 
+
 #
 # Submit Trait
 #
 @data_bp.route('/trait/start-submit')
 @cache.memoize(60*60)
 @jwt_required()
+@check_feature_flag('PHENOTYPE_DB_ENABLED')
```
r-vieira, line 91: 👍

src/modules/site-v2/base/views/data/data.py
```diff
@@ -132,97 +147,147 @@ def submit_trait_form():
...
-  phenotype_values = get_phenotype_values_for_trait(id)
+@check_feature_flag('PHENOTYPE_DB_ENABLED')
+@parse_trait(validate_owner=True, allow_admin=True)
+def trait(trait: Trait):
+  """
+    View / Review Trait Page
+  """
+
+  # Track which endpoint was called
+  reviewing = request.endpoint.endswith('review_trait')
```
r-vieira, line 168: style detail: consider avoiding "double route handling", if using blueprint routes, consider keeping each route separate by blueprint, and if needed, decouple common business logic into reusable component/helper/etc.

src/modules/site-v2/base/views/api/trait.py
```diff
@@ -301,16 +310,16 @@ def query_list_datatable(user_filter=None, status_filter=None):
...
+  #   except:
+  #     abort(422, description="Invalid search")
 
     # Treat non-dict values as search strings
     # Use the original raw string here so JSON casting doesn't change the value
     # (e.g. JSON "true" becoming Python "True")
-    if not isinstance(search_full, dict):
-      search_full = { 'search_val': search_raw }
+    # if not isinstance(search_full, dict):
+    search_full = { 'search_val': search_raw }
```
r-vieira, line 322: double check all the code paths and confirm that there is no path to a XSS (rendering unescaped user input as highlighted by gh's code scanning)

r-vieira (changes requested):
looks great! Double check the XSS issue (might be false-positive)

r-vieira (changes requested):
@orzut @vlagrassa  -- please review the code scanning warning below and confirm that all code-paths yield sanitized user input

r-vieira (approved):
adding the XSS false-positive as a task

PR #496, opened 2024-07-11
Minor Updates

vlagrassa:


src/modules/site-v2/base/views/api/trait.py
```diff
@@ -71,13 +71,19 @@ def matches_any(cls, endpoint, endpoint_type_set):
 def filter_trait_files(tf):
   return tf.is_public and not tf.is_bulk_file
 
+def _clean_recursive(obj):
+  if isinstance(obj, str):
+    return bleach.clean(obj)
+  elif isinstance(obj, list):
+    return [ _clean_recursive(sub_obj) for sub_obj in obj ]
```
r-vieira, line 78: Unlikely that this will happen with good inputs, but be careful while handling cycles in the lists and running into max recursion depth, eg:
```python
a = list()
a.append(1)
a.append(a)
_clean_recursive(a)
```
will throw a `RecursionError: maximum recursion depth exceeded while calling a Python object`

r-vieira (approved):
looks good.

PR #497, opened 2024-07-11
Indel Finder "Rerun" Option

vlagrassa:
- On admin-only "View All Reports" page, add an optional column to re-run a job that ended in an error
- Enable Rerun column for Indel Finder (with an admin-only "resubmit" endpoint)

In the future, if an Indel Finder job fails for a user, we'll now be able to easily re-run their report once we've fixed the container / site.

TODO: Should we enable this for the other tools? Should discuss the way their data is stored.  Indel Finder lends itself better to this, since the user doesn't provide any upload file, they just choose one of our pre-defined regions.


r-vieira:
> TODO: Should we enable this for the other tools? Should discuss the way their data is stored. Indel Finder lends itself better to this, since the user doesn't provide any upload file, they just choose one of our pre-defined regions.
-- yes, let's discuss details.

r-vieira (approved):
looks good!

PR #499, opened 2024-07-11
Genome Browser Tracks Admin Page & Safer Lookups

vlagrassa:
- Move genome browser track endpoints to their own path within the API
- Reference browser tracks by their GCP unique ID, rather than their display name
- Update Genome Browser admin page(s) to display list of browser tracks and allow creating / editing
  -  Uses a re-orderable list, like Announcements page, to set order of tracks on Genome Browser page
  - Lets us choose which track(s) show up in each tool browser
  - Displays file locations in GCP

NOTE: Before deploying to QA / main, I have to update the browser track objects in GCP datastore.  Luckily, this change is backwards-compatible, so I can make the change first, we can re-deploy, then I can clean up old field(s).


src/modules/site-v2/base/views/api/strains.py
```diff
@@ -0,0 +1,37 @@
...
+@jwt_required()
+def get_sv_strains():
+
+  # Helper function to try getting the strain list for a species and print an error message if not available
+  def try_get_sv_strains(species):
+    try:
+      return get_sv_strains_helper(species)
+    except Exception as e:
+      logger.error(f"Couldn't find strain variant annotations for {species}. Make sure the appropriate VCF file exists. Full error: {e}")
+      return []
```
r-vieira, line 32: Ok to log the error in the exception handler, but should it bubble up instead of returning `[]`? -- will the caller know the difference between `[]` and `[]` from an exception?

src/modules/site-v2/base/views/api/tracks.py
```diff
@@ -0,0 +1,285 @@
...
+    })
+    new_track.save()
+
+    # Try explicitly placing the new track at the bottom of the list
+    # If this fails, it will get a default order value, so we can ignore errors
+    try:
+      new_track['order'] = len(BrowserTrackDefault.query_ds())
+      new_track.save()
+    except Exception as ex:
+      pass
```
r-vieira, line 158: should we log the `ex` here?

src/modules/site-v2/templates/admin/browser_tracks/edit.html
```diff
@@ -0,0 +1,210 @@
...
+      e.preventDefault();
+      window.location = "{{ url_for('admin_gene_browser_tracks.admin_gene_browser_tracks') }}";
+    });
+
+    $(".multi-checkbox input").addClass("form-check-input mt-0");
+
+    // Prevent enter key from submitting form inside Markdown editor
+    // Primarily affects hyperlink insert popup
+    $('#{{ form.information.id }}').keydown(function(event) {
+      if (event.keyCode == 13) {
```
r-vieira, line 201: always prefer `===` instead of `==`

src/modules/site-v2/templates/tools/genome_browser/gbrowser.html
```diff
@@ -254,16 +251,82 @@ <h1 class="modal-title fs-5" id="infoModalLabel">Track Information</h1>
...
+          // If there's no description info to display, continue to the next track
+          if (!parsed_data[track_name].information) return;
+
+          // Section header (track name)
+          const header = document.createElement('h2');
+          header.classList.add('h4');
+          header.innerText = track_name;
+
+          // Section body (track information / description)
+          const body = document.createElement('div')
```
r-vieira, line 290: semantically we may want to avoid `body` since `body` exists at the top level.

src/pkg/caendr/caendr/models/datastore/dataset_release.py
```diff
@@ -150,6 +150,54 @@ def browser_tracks(self, val):
...
+
+      Returns `True` if the operation changed this object,
+      i.e. if the track was initially supported by this release.
+    '''
+    if track['display_name'] in self['browser_tracks']:
+      self['browser_tracks'].remove(track['display_name'])
+      return True
+    return False
+
+  def set_browser_track(self, track, value: bool) -> bool:
```
r-vieira, line 188: the function signature could signal the operation (mainly around naming); instead of `set_` consider `modify_` and instead of `value` use `is_visible` or similar to convey type.

r-vieira (approved):
looks great! See minor inline comments

PR #509, opened 2024-08-25
Feature/download all traits

orzut:
- execute query by chunks
- set Gunicorn timeout to 4800 seconds


src/modules/site-v2/Dockerfile
```diff
@@ -63,4 +63,4 @@ RUN echo "\n\n# GIT_COMMIT automatically added during the makefile build contain
 RUN echo "GIT_COMMIT=$GIT_COMMIT" >> module.env
 
 CMD export $(cat .env | sed -e '/^\s*$/d' | sed -e '/^\#/d' | xargs) && \
-  gunicorn -b :$PORT main:app
+  gunicorn -b :$PORT --timeout=4800 main:app
```
r-vieira, line 66: can you confirm that `4800` is within the hard limits of `GCP`?

r-vieira (approved):
approved. See inline comments.

PR #510, opened 2024-09-04
Update yield number from 1000 to 100

orzut:
Updated yield number for query execution, as there were 'request timed out' warnings in the production site. Already tested out with `yield_per(100)` -> all rows have been successfully written to CSV.

### DIFF
--- a/src/modules/site-v2/base/views/tools/phenotype_database.py
+++ b/src/modules/site-v2/base/views/tools/phenotype_database.py
@@ -92,7 +92,7 @@ def download_csv():
   def generate():
     yield file_format['sep'].join(columns) + '\n'
     try:
-      for row in traits.yield_per(1000):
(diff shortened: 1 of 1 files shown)


src/modules/site-v2/base/views/tools/phenotype_database.py
```diff
@@ -92,7 +92,7 @@ def download_csv():
   def generate():
     yield file_format['sep'].join(columns) + '\n'
     try:
-      for row in traits.yield_per(1000):
+      for row in traits.yield_per(100):
```
r-vieira, line 95: avoid magic numbers -- consider creating a variable that describes the meaning of `100`, e.g.: `TRAITS_PER_CHUNK=100`, etc.

r-vieira (approved):
looks good! See inline comments

